PR #2, opened 2019-11-22
bootstrap client and user service

lieut-data:
This commit only solves the problem that is `*model.AppError`, but otherwise proxies through the underlying API unchanged at this time.

### DIFF
--- a/user.go
+++ b/user.go
@@ -1,0 +1,83 @@
+package pluginapi
+
+import (
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mattermost-server/v5/plugin"
+)
+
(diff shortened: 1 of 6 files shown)


user.go
```diff
@@ -0,0 +1,83 @@
...
+func (u *UserService) CreateUser(user *model.User) (*model.User, error) {
+	user, appErr := u.api.CreateUser(user)
+
+	return user, normalizeAppErr(appErr)
+}
+
+// DeleteUser deletes a user.
+//
+// Minimum server version: 5.2
+func (u *UserService) DeleteUser(userID string) error {
```
iomodo: `ID` made my day <3

PR #3, opened 2019-11-23
add KV service

ilgooz:


kv.go
```diff
@@ -0,0 +1,143 @@
...
+	api plugin.API
+}
+
+// SetOption used to configure a Set() operation.
+type SetOption func(*model.PluginKVSetOptions)
+
+// SetCompareCurrentValueOption returns an option to conditionally update a key's
+// value by comparing given value to match with the currently set one.
+// this is an atomic operation.
+func SetCompareCurrentValueOption(value interface{}) SetOption {
```
lieut-data: Propose small nomenclature change:
```
// SetAtomic guarantees the write will occur only when the current value of matches the given old value. A client is expected to read the old value first, then pass it back to ensure the value has not since been modified.
func SetAtomic(oldValue interface{}) SetOption {
```

lieut-data:
Super work, @ilgooz! I've made one comment above, but not going to block things -- merging!

PR #5, opened 2019-11-23
added File service

ilgooz:


file.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// FileService provides features to deal with post attachments.
+type FileService struct {
+	api plugin.API
+}
+
+// Get gets content of a file by id.
+//
+// @tag File minimum server version: 5.8
```
lieut-data: We can probably eliminate all the `@tag` stuff, now that the grouping is obvious given the service.

file.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Upload uploads a file to a channel to be later attached to a post.
+//
+// @tag File @tag Channel minimum server version: 5.6
+func (f *FileService) Upload(content io.Reader, fileName, channelID string) (*model.FileInfo, error) {
+	contentBytes, err := ioutil.ReadAll(content)
+	if err != nil {
+		return nil, err
+	}
+	info, aerr := f.api.UploadFile(contentBytes, channelID, fileName)
```
lieut-data: nit: suggest `appErr` for consistency with other Mattermost code.

PR #27, opened 2019-11-27
added dlock (distributed lock) -alpha

ilgooz:
Hey, here is the initial version of dlock. Thanks!

closes https://github.com/lieut-data/mattermost-plugin-api/issues/20

### DIFF
--- a/dlock/dlock.go
+++ b/dlock/dlock.go
@@ -121,7 +121,8 @@ func (d *DLock) startRefreshLoop() {
 			case <-t.C:
 				_, err := d.api.Set(d.key, true, pluginapi.SetExpiry(lockTTL))
 				if err != nil {
-					d.api.Error(errors.Wrap(err, "cannot refresh a lock, Store.Set() returned with an error").Error())
+					err = errors.Wrap(err, "cannot refresh a lock, Store.Set() returned with an error")
+					d.api.Error(err.Error())
 				}
 			case <-ctx.Done():
 				return
@@ -137,8 +138,10 @@ func (d *DLock) startRefreshLoop() {
 func (d *DLock) Unlock() error {
 	d.refreshCancel()
 	d.refreshWait.Wait()
-	_, err := d.api.Set(d.key, nil)
-	return err
+	if _, err := d.api.Set(d.key, nil); err != nil {
+		return errors.Wrap(err, "cannot release a lock, Store.Set() returned with an error")
+	}
+	return nil
 }
 
 // buildKey builds a lock key for KV store.
--- a/dlock/dlock.go
+++ b/dlock/dlock.go
@@ -31,15 +31,16 @@ const (
 	lockTryInterval = time.Second
 )
 
-// Store is a data store to keep locks' state.
-type Store interface {
+// API is a portion of pluginapi to keep locks' state and log errors.
+type API interface {
 	Set(key string, value interface{}, options ...pluginapi.KVSetOption) (bool, error)
+	Error(message string, keyValuePairs ...interface{})
 }
 
 // DLock is a distributed lock.
 type DLock struct {
-	// store used to store lock's state to do synchronization.
-	store Store
+	// api used to keep locks' state and log errors.
+	api API
 
 	// key to lock for.
 	key string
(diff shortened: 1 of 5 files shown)


dlock/dlock.go
```diff
@@ -0,0 +1,150 @@
...
+	// lockRefreshInterval used to determine how long to wait before refreshing
+	// a lock's expiry time.
+	lockRefreshInterval = time.Second
+
+	// lockTryInterval used to wait before trying to obtain the lock again.
+	lockTryInterval = time.Second
+)
+
+// API is a portion of pluginapi to keep locks' state and log errors.
+type API interface {
```
ilgooz: I think it's best to split this interface into two, as Logger and Store and accept them separately at New(). That way it's easier for caller to pass necessary implementations and to avoid possible name conflicts inside possible big variations of API interfaces. I suggest this as an idiomatic pattern for plugins.

This can be accomplished in a separate PR.

dlock/dlock.go
```diff
@@ -0,0 +1,150 @@
...
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-afterC:
+		}
+	}
+}
+
+// TryLock tries to obtain the lock immediately.
+// err is only filled on system failure.
+func (d *DLock) TryLock() (isLockObtained bool, err error) {
```
hanzei: 1/5 In general I don't see the necessity to name the return values. The go doc comment should make clear, what the method does.

dlock/dlock.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	if isLockObtained {
+		d.startRefreshLoop()
+	}
+	return isLockObtained, nil
+}
+
+// startRefreshLoop refreshes an obtained lock to not get caught by lock's TTL.
+// TTL tends to hit and release the lock automatically when plugin terminates.
+func (d *DLock) startRefreshLoop() {
```
hanzei: Clever way to do this :+1:

dlock/dlock.go
```diff
@@ -0,0 +1,150 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		t := time.NewTicker(lockRefreshInterval)
```
hanzei: Maybe try to get the lock a little bit before the TTL expires to avoid races?

dlock/dlock.go
```diff
@@ -0,0 +1,150 @@
...
+	}()
+	d.refreshCancel = cancel
+	d.refreshWait = &wg
+}
+
+// Unlock unlocks Lock().
+// use Unlock() exactly like sync.Mutex.Unlock().
+func (d *DLock) Unlock() error {
+	d.refreshCancel()
+	d.refreshWait.Wait()
```
hanzei: Can you elaborate why we need wait here?

dlock/dlock_test.go
```diff
@@ -0,0 +1,48 @@
+package dlock
```
hanzei: Can this be in `dlock_test` instead?

dlock/dlock_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+func TestLock(t *testing.T) {
+	dl := New("a", dlocktest.New())
+	var wg sync.WaitGroup
+	wg.Add(3)
+	for i := 0; i < 3; i++ {
+		go func() {
+			defer wg.Done()
+			for i := 0; i < 10; i++ {
+				dl.Lock(context.Background())
```
hanzei: Can you please check all the returned errors?

dlock/dlock_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/lieut-data/mattermost-plugin-api/dlock/dlocktest"
+	"github.com/stretchr/testify/require"
+)
+
+// TODO(ilgooz): test all branches including related ones to Store errors and ExpireInSeconds.
+// TODO(ilgooz): can move tests from sync/mutex_test.go.
+
+func TestLock(t *testing.T) {
+	dl := New("a", dlocktest.New())
```
hanzei: Maybe it makes sense to also test dlock with a mocked API instead of you in-memory implementation to make sure the API calls work as expected.

dlock/dlock_test.go
hanzei: Maybe alter the test to also check the rough order in which the locks are acquired.

lieut-data:
@ilgooz / @hanzei, I'm going to go ahead and merge this as-is, allowing us to iterate on the alpha efforts here and eventually unblock our larger goal of promoting this repository to the `mattermost` organization.

PR #28, opened 2019-12-10
Fix kv set and move back to upstream

hanzei:
- No more custom fork
- KVSet is synced up with the fix on MM `master`

cc @ilgooz

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ import (
 
 Finally, add this package as a dependency:
 ```sh
-go get github.com/lieut-data/mattermost-plugin-api@latest
+go get github.com/lieut-data/mattermost-plugin-api
 ```
 
 ## Migration Guide
--- a/README.md
+++ b/README.md
@@ -26,14 +26,9 @@ import (
 )
 ```
 
-Then update your `go.mod` to point at a custom fork of the `mattermost-server`. This requirement is temporary, as the changes will soon be available upstream:
-```sh
-go mod edit -replace github.com/mattermost/mattermost-server/v5=github.com/lieut-data/mattermost-server/v5@v5.99.99
-```
-
 Finally, add this package as a dependency:
 ```sh
-go get github.com/lieut-data/mattermost-plugin-api@v0.0.3
+go get github.com/lieut-data/mattermost-plugin-api@latest
 ```
 
 ## Migration Guide
--- a/go.mod
+++ b/go.mod
@@ -3,10 +3,8 @@ module github.com/lieut-data/mattermost-plugin-api
 go 1.13
 
 require (
-	github.com/mattermost/mattermost-server/v5 v5.99.99
+	github.com/mattermost/mattermost-server/v5 v5.3.2-0.20191203094615-d0d6ce0a7014
 	github.com/pkg/errors v0.8.1
 	github.com/sirupsen/logrus v1.4.2
 	github.com/stretchr/testify v1.4.0
 )
-
-replace github.com/mattermost/mattermost-server/v5 => github.com/lieut-data/mattermost-server/v5 v5.99.99
--- a/kv.go
+++ b/kv.go
@@ -17,35 +17,28 @@ type KVService struct {
 	api plugin.API
 }
 
+type kvSetOptions struct {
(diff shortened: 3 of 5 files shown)


kv_test.go
```diff
@@ -235,9 +229,7 @@ func TestDelete(t *testing.T) {
 	defer api.AssertExpectations(t)
 	client := pluginapi.NewClient(api)
 
-	api.On("KVSetWithOptions", "1", nil, model.PluginKVSetOptions{
-		EncodeJSON: true,
-	}).Return(true, nil)
+	api.On("KVSetWithOptions", "1", []byte(nil), model.PluginKVSetOptions{}).Return(true, nil)
```
hanzei: I thought using `nil` instead of `[]byte(nil)` should be fine here, but it isn't. Maybe a bug in `testify`?

README.md
```diff
@@ -26,14 +26,9 @@ import (
...
 
-Then update your `go.mod` to point at a custom fork of the `mattermost-server`. This requirement is temporary, as the changes will soon be available upstream:
-```sh
-go mod edit -replace github.com/mattermost/mattermost-server/v5=github.com/lieut-data/mattermost-server/v5@v5.99.99
-```
-
 Finally, add this package as a dependency:
 ```sh
-go get github.com/lieut-data/mattermost-plugin-api@v0.0.3
+go get github.com/lieut-data/mattermost-plugin-api@latest
```
lieut-data: Is the `latest` strictly necessary? Do we just need to add a new tag after this commit?

kv.go
```diff
@@ -18,22 +18,36 @@ type KVService struct {
...
 // value. A client is expected to read the old value first, then pass it back to ensure the value
 // has not since been modified.
 func SetAtomic(oldValue interface{}) KVSetOption {
-	return func(o *model.PluginKVSetOptions) {
+	return func(o *model.PluginKVSetOptions) error {
 		o.Atomic = true
-		o.OldValue = oldValue
+		if oldValue != nil {
+			oldValueBytes, isOldValueInBytes := oldValue.([]byte)
+			if isOldValueInBytes {
```
lieut-data: This /seems/ like the wrong place to do this -- why not just check in the main call instead?

README.md
hanzei: Yes, we need to add a new tag. No, the `latest` is not strictly needed. Will remove.

kv.go
hanzei: If I move the code to the main call, I would have to store `oldValue` somewhere, but `PluginKVSetOptions` only has a `OldValue []byte`.

go.mod
```diff
@@ -3,10 +3,8 @@ module github.com/lieut-data/mattermost-plugin-api
 go 1.13
 
 require (
-	github.com/mattermost/mattermost-server/v5 v5.99.99
+	github.com/mattermost/mattermost-server/v5 v5.3.2-0.20191203094615-d0d6ce0a7014
```
lieut-data: To double check: is this `v5.3.2` expected?

kv.go
lieut-data: Could we use an unexported struct embedding and extending `model.PluginKVSetOptions` instead? I'd love to keep the business logic here as closely coupled together as possible, vs. splitting it between here and the main call.

go.mod
hanzei: Yes it is. When using a specific commit (which is needed because there no releases contains these changes), go uses `v5.3.2` as a prefix. Don't ask me why :man_shrugging:

kv.go
hanzei: I applied your suggestion in c570432d10ce8044b632ad56caf3b1af4922044b.

While I see the point keeping the logic close together, with this change the coupling between the downstream opts and the opts struct used here is lost. Maybe this is a good thing, because new options downstream are not exposed to the public immediately, but it make the code harder to read and understand.

I'm fine merging this PR with the changes, but want to keep the discussion on how close the downstream code could be coupled going.

lieut-data:
Sounds good, @hanzei -- going to go ahead and merge this, and we can iterate later as necessary.

PR #33, opened 2020-02-26
Cluster user-space key expiry

lieut-data:
This changes the implementation of the `cluster` package's Mutex to eschew `KVSetWithOptions` and the corresponding server-support for expiring keys. Instead, those semantics are moved to "user-space", with the mutex itself handling key expiry.

The rationale for this change is to enable use of this package for v5.16 and later instead of requiring the yet-to-be-released v5.21 that [patches](https://github.com/mattermost/mattermost-server/pull/13858) handling of atomic key value expiry. This makes the package immediately useful to ongoing plugin development.

Note that we are still consuming the raw RPC API, and nothing from the parent repository.

### DIFF
--- a/cluster/mutex.go
+++ b/cluster/mutex.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"strconv"
 	"sync"
 	"time"
 
@@ -24,6 +25,7 @@ const (
 
 // MutexPluginAPI is the plugin API interface required to manage mutexes.
 type MutexPluginAPI interface {
+	KVGet(key string) ([]byte, *model.AppError)
 	KVSetWithOptions(key string, value []byte, options model.PluginKVSetOptions) (bool, *model.AppError)
 	LogError(msg string, keyValuePairs ...interface{})
 }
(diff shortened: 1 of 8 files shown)


cluster/mutex.go
```diff
@@ -58,47 +52,109 @@ type Mutex struct {
...
 	return &Mutex{
 		pluginAPI: pluginAPI,
-		key:       key,
+		key:       makeLockKey(key),
 	}
 }
 
+// makeLockKey returns the prefixed key used to namespace mutex keys.
+func makeLockKey(key string) string {
+	if len(key) == 0 {
```
levb: 0/5 return a "" which is bound to fail downstream. `panic` in production code seems like something to avoid?

lieut-data: My original rationale for panicking here was to make the `NewMutex` constructor not have to return an `error`, effectively simplifying the interface and ultimately making it just a little more accessible (and more like `sync.Mutex`). But your comment reminded me to vet this thinking, and I came across this clear directive from https://blog.golang.org/defer-panic-and-recover:

> The convention in the Go libraries is that even when a package uses panic internally, its external API still presents explicit error return values.

I'll change this back -- thank you!

lieut-data:
After discussion with @aaronrothschild, it's agreed that we want to require v5.21+ to use the `cluster` package, making these changes no longer required.

PR #35, opened 2020-03-02
cluster: support LockWithContext

lieut-data:
#### Summary
This branches off https://github.com/mattermost/mattermost-plugin-api/pull/33 to add support for locking with a context (and thus implicitly supporting lock cancellation, "try lock" semantics, etc.).

I opted not to merge directly to `master` given the rework happening in that active PR.

### DIFF
--- a/cluster/mutex.go
+++ b/cluster/mutex.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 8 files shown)


cluster/mutex.go
```diff
@@ -161,10 +162,23 @@ func (m *Mutex) refreshLock() error {
 // Lock locks m. If the mutex is already locked by any plugin instance, including the current one,
 // the calling goroutine blocks until the mutex can be locked.
 func (m *Mutex) Lock() {
+	m.LockWithContext(context.Background())
+}
+
+// Lock locks m unless the context is cancelled. If the mutex is already locked by any plugin
```
mickmister: Nit: function name in comment
```suggestion
// LockWithContext locks m unless the context is cancelled. If the mutex is already locked by any plugin
```

lieut-data:
@ali-farooq0, is this something you can take a look at soon?

ali-farooq0:
oh, the magical command doesn't work here ;/

PR #38, opened 2020-03-11
Support externally defined wait intervals

lieut-data:
#### Summary
Parameterize the function for computing a job's next wait interval, and provide two implementations by default:
* `MakeWaitForInterval`: the existing semantics of spacing runs by a configured interval
* `MakeWaitForRoundedInterval`: a variation of the above that rounds (really, truncates) to that interval, so as to consistently run, say, every 5 minutes on 0:00, 0:05, 0:10, ... even if the job is paused for some reason. In other words, it won't drift out of sync relative to the clock. It doesn't guarantee /never/ running less than the interval, since if it's been longer than the required interval it just runs right away.

@mickmister, hopefully this addresses your needs with the MS calendar plugin: but if not, hopefully you can provide your own wait interval function. Let me know if you think the existing algorithm can/should be tweaked.

Also, apologies for the nested pull requests here -- I'll rebase as necessary, but these are all building off each other. I'll clean this up as we merge in.

### DIFF
--- a/cluster/job.go
+++ b/cluster/job.go
@@ -27,17 +27,73 @@ type JobConfig struct {
 	Interval time.Duration
 }
 
+// NextWaitInterval is a callback computing the next wait interval for a job.
+type NextWaitInterval func(now time.Time, metadata jobMetadata) time.Duration
+
+// MakeWaitForInterval creates a function to scheduling a job to run on the given interval relative
+// to the last finished timestamp.
+//
+// For example, if the job first starts at 12:01 PM, and is configured with interval 5 minutes,
+// it will next run at:
+//     12:06, 12:11, 12:16, ...
+//
+// If the job has not previously started, it will run immediately.
+func MakeWaitForInterval(interval time.Duration) NextWaitInterval {
+	if interval == 0 {
+		panic("must specify non-zero ready interval")
+	}
+
+	return func(now time.Time, metadata jobMetadata) time.Duration {
+		sinceLastFinished := now.Sub(metadata.LastFinished)
+		if sinceLastFinished < interval {
+			return interval - sinceLastFinished
+		}
+
+		return 0
+	}
+}
+
+// MakeWaitForRoundedInterval creates a function, scheduling a job to run on the nearest rounded
+// interval relative to the last finished timestamp.
+//
+// For example, if the job first starts at 12:04 PM, and is configured with interval 5 minutes,
+// and is configured to round to 5 minute intervals, it will next run at:
+//	12:05 PM, 12:10 PM, 12:15 PM, ...
+//
+// If the job has not previously started, it will run immediately. Note that this wait interval
+// strategy does not guarantee a minimum interval between runs, only that subsequent runs will be
+// scheduled on the rounded interval.
+func MakeWaitForRoundedInterval(interval time.Duration) NextWaitInterval {
+	if interval == 0 {
+		panic("must specify non-zero ready interval")
+	}
+
+	return func(now time.Time, metadata jobMetadata) time.Duration {
+		if metadata.LastFinished.IsZero() {
+			return 0
+		}
+
+		target := metadata.LastFinished.Add(interval).Truncate(interval)
+		untilTarget := target.Sub(now)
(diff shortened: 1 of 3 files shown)


cluster/job_test.go
```diff
@@ -11,23 +11,268 @@ import (
...
+		t.Run(testCase.Description, func(t *testing.T) {
+			now := time.Now()
 
-		mockPluginAPI := newMockPluginAPI(t)
+			var lastFinished time.Time
+			if testCase.LastFinished != neverRun {
+				lastFinished = now.Add(testCase.LastFinished)
+			}
+
+			actual := MakeWaitForInterval(testCase.Interval)(now, jobMetadata{
```
mickmister: Comment for off-topic discussion - This function takes the value of `Now` which makes it safe to use as a point of reference for testing. I'm not sure if the `cluster` package ever uses `time.Now()` in its source, but I thought it may be beneficial to mention the package https://github.com/benbjohnson/clock, used to manipulate the "system's" `Now` time during testing.

I've made a function `var nowFunc = time.Now` in my source, then used the `clock` package in the test, temporarily assigning the `nowFunc` to the clocks `Now` function. Depending on the amount of time-based functionality in the `cluster` package, it may be beneficial to introduce something like `clock` in the tests. I'm curious what your thoughts are on this @lieut-data

cluster/job.go
```diff
@@ -27,17 +27,73 @@ type JobConfig struct {
...
+	if interval == 0 {
+		panic("must specify non-zero ready interval")
+	}
+
+	return func(now time.Time, metadata jobMetadata) time.Duration {
+		if metadata.LastFinished.IsZero() {
+			return 0
+		}
+
+		target := metadata.LastFinished.Add(interval).Truncate(interval)
```
mickmister: Great way to handle this! If I understand this correctly, this is how an example situation would play out:

- Rounded interval is set to `5 minutes`
- Job starts at `12:04PM`
- First run runs at `12:05PM`
- The job lasts `7 minutes`
- The job will not run at `12:10PM`, and will instead run next at `12:15PM`

A job meant to run every 5 minutes should not last that long (of course it could), but I just want to make sure I understand the functionality.

cluster/job_test.go
lieut-data: Ooh, that's a very slick approach. It's always a pain mocking the value of `time.Now()`, since it never feels like a factorization that improves the code: just something awkward that's necessary. I'll make a point of (asynchronously) revisiting the clock package you mention and seeing if things can be simpler as a result. Thank you!!

cluster/job.go
lieut-data: That's (almost) correct -- we anchor from the time the job last finished to determine when to next run the job. Since the job would have finished at `12:11PM`, it would indeed next run at `12:15PM`.

The nuance is that the "first run" will actually occur at `12:04PM`, with the second run occurring at `12:05PM` and subsequent runs then being properly spaced out. I opted for this edge case handling because it eliminated the need to supply a `startTime` from which to anchor the calculations. Without an explicit `startTime` (i.e. when there is no `metadata.LastFinished` yet), it's not obvious how to wait for the top of the interval. By the time we reach it, we're in a new interval, and might decide to wait for the top of /that/ one. 

I suppose we could tweak this so that `12:00:00-12:02:29` would round down and `12:02:30-12:04:49` would round up. My assumption was that these jobs would likely be "idempotent" and that running slightly more often that necessary in one case would be worth the simpler API surface and implementation.

levb: I like the current implementation's consistency, 1/5 against special-cased round-down

cluster/job_test.go
levb: 0/5 I had the exact same problem in Solar Lottery and decided to stick with the less slick, intention-revealing `now` parameter. I'd be ok going either way, but we should try to be consistent.

mickmister: I thought about this, and using the `now` parameter makes much more sense that assigning a `nowFunc` at package level. Using the `clock` package in the test, while using the `now` function parameter, gives the test a flexible way to provide the dependency.

Issue #17, opened 2019-11-27
Add GetManifest method


hanzei:
See https://github.com/mattermost/mattermost-plugin-starter-template/pull/73#issuecomment-557272595

PR #42, opened 2020-04-03
Add GetManifest

hanzei:
#### Summary
Add `GetManifest() (*model.Manifest, error)` to `SystemService`.

I did add caching for the manifest as it should never be altered.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-api/issues/17


client.go
```diff
@@ -32,22 +32,22 @@ func NewClient(api plugin.API) *Client {
...
+		File:          FileService{api: api},
+		Frontend:      FrontendService{api: api},
+		Group:         GroupService{api: api},
+		KV:            KVService{api: api},
+		Log:           LogService{api: api},
+		Mail:          MailService{api: api},
+		Plugin:        PluginService{api: api},
+		Post:          PostService{api: api},
+		Session:       SessionService{api: api},
+		System:        SystemService{api: api},
```
hanzei, line 49: I did alter this to avoid having to pass `nil` for `*model.Manifest`

system.go
```diff
@@ -5,11 +5,41 @@ import (
...
+
+// GetManifest returns the manifest from the plugin bundle.
+//
+// Minimum server version: 5.10
+func (s *SystemService) GetManifest() (*model.Manifest, error) {
+	// Use a copy of cache manifest is present
+	if s.m != nil {
+		m := *s.m
+		return &m, nil
+	}
```
hanzei, line 25: Is there a more idiomatic way to do this?

system.go
```diff
@@ -5,11 +5,41 @@ import (
 
 	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/mattermost/mattermost-server/v5/plugin"
+	"github.com/pkg/errors"
 )
 
 // SystemService exposes methods to query system properties.
 type SystemService struct {
 	api plugin.API
+	m   *model.Manifest
```
hanzei, line 14: Maybe there should be a lock to ensure the method is safe thread

system.go
```diff
@@ -5,11 +5,41 @@ import (
...
+
+	m, _, err := model.FindManifest(path)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to find and open manifest")
+	}
+
+	s.m = m
+
+	// Copy before returning
+	rm := *m
```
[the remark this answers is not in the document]
hanzei: Makes sense. Maybe I was over over engineering.

client.go
lieut-data: Can we revert these changes now that we're not caching?

hanzei: We could, but I would like to keep them as they make it clear what is used to create the services.

hanzei:
@lieut-data With this method in place how do you feel about dropping `server/manifest.go` from the starter template and just telling people to use this method?

lieut-data:
Ah, I'd forgotten about the `manifest.go` stuff altogether. I think my comment at https://github.com/mattermost/mattermost-plugin-starter-template/pull/73#issuecomment-557277915 still applies... and so, actually I'm questioning what drove this addition? Reading your own "source code" back from an RPC API remains counterintuitive to me.

hanzei:
> Reading your own "source code" back from an RPC API remains counterintuitive to me.

Maybe, but that is for users of this library hidden. From a code perspective it seams seam fine to me.

`make apply` has caused issues in the past with e.g. people forgetting to run it or conflicts happening in multiple places.

lieut-data:
Ah, interesting. Thanks for the context!

PR #45, opened 2020-04-22
MM-24456 - CreateChannel followed by AddChannelMember is racy when a replica exists

ali-farooq0:
#### Summary
Calling `pluginAPI.AddUserToChannel(...)` can lead to a race condition if it is proceeded by `CreateChannel` as `AddUserToChannel` first calls [`api.GetChannel`](https://github.com/mattermost/mattermost-server/blob/aa47b4633f18eed917c70263d139376f15d8778c/app/plugin_api.go#L422) to validate channel before adding the user. `GetChannel` uses a replica if one exists and if data hasn't reached consistency, this call returns a `channel not found error`. Solution is to wait for channel creation to propagate through replicas before returning success from `Channel.Create(..)`

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-24456

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -101,7 +101,7 @@ func (c *ChannelService) Create(channel *model.Channel) error {
 
 	*channel = *createdChannel
 
-	return c.waitForChannelCreation(createdChannel.Id)
+	return c.waitForChannelCreation(channel.Id)
 }
 
 // Update updates a channel.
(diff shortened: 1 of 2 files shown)


channel.go
```diff
@@ -213,6 +217,23 @@ func (c *ChannelService) UpdateChannelMemberNotifications(channelID, userID stri
 	return channelMember, normalizeAppErr(appErr)
 }
 
+func (c *ChannelService) waitForChannelCreation(channelID string) error {
+	now := time.Now()
```
lieut-data, line 221: Can we additionally gate this on a check for a replica, like the server code does?

ali-farooq0: Oh whoops forgot we had access to the config. Check added.

PR #49, opened 2020-06-12
MM-25017 - add SetAtomicWithRetries

cpoile:
#### Summary
- Simplifies a compare and set operation with a certain amount of retry attempts
- Related to: https://github.com/mattermost/mattermost-plugin-incident-response/pull/155

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-24017


kv.go
```diff
@@ -138,6 +139,45 @@ func (k *KVService) CompareAndDelete(key string, oldValue interface{}) (bool, er
...
+// `valueFunc` is a user-provided function that will take the old value and return the new value
+//    or an error.
+// `numRetries is the number of times the function will attempt to compare and set before failing.
+//
+// Returns err if the key could not be retrieved (DB error), valueFunc returned an error,
+//   if the key could not be set (DB error), or if the key could not be set (after retries)
+// Returns nil if the value was set
+//
+// Minimum server version: 5.18
+func (k *KVService) SetAtomicWithRetries(key string, valueFunc func(oldValue interface{}) (newValue interface{}, err error), numRetries int) error {
```
crspeller, line 157: Does numRetries need to be an option? It doesn't seem situational. Maybe we should just be opinionated about it.

cpoile: Yah, that makes sense. Will do.

kv_test.go
```diff
@@ -195,6 +196,123 @@ func TestCompareAndDelete(t *testing.T) {
...
+		key       string
+		valueFunc func(oldValue interface{}) (newValue interface{}, err error)
+		setupAPI  func(api *plugintest.API)
+		wantErr   bool
+	}{
+		{
+			name: "Test SetAtomicWithRetries success after first attempt",
+			key:  "testNum",
+			valueFunc: func(old interface{}) (interface{}, error) {
+				return 2, nil
```
ali-farooq0, line 211: Should we assert that the `old` value here is actually `oldJSONBytes` because it's possible that the func is passing in wrong `old` value and that these tests are still passing? thoughts?

kv_test.go
```diff
@@ -195,6 +196,123 @@ func TestCompareAndDelete(t *testing.T) {
...
+					Atomic:   true,
+					OldValue: oldJSONBytes,
+				}).Return(true, nil).Once()
+			},
+		},
+		{
+			name: "Test SetAtomicWithRetries failure on get",
+			key:  "testNum",
+			valueFunc: func(old interface{}) (interface{}, error) {
+				return 2, nil
```
ali-farooq0, line 247: should we assert in this case that this func was not called?

kv_test.go
```diff
@@ -195,6 +196,123 @@ func TestCompareAndDelete(t *testing.T) {
...
+				api.On("KVGet", "testNum").Return(oldJSONBytes, nil).Once()
+				api.On("KVSetWithOptions", "testNum", newJSONBytes, model.PluginKVSetOptions{
+					Atomic:   true,
+					OldValue: oldJSONBytes,
+				}).Return(false, newAppError()).Once()
+			},
+			wantErr: true,
+		},
+		{
+			name: "Test SetAtomicWithRetries failure on five set attempts -- depends on numRetries constant being = 5",
```
ali-farooq0, line 284: Great tests!! Could I ask you to write one more where each `KVGet` Returns a different value and `KVSetWithOptions` expects a different value per iteration (with at least 3-4 iterations or maybe a max of 5).

As I think it is possible if you were to change the implementation to always pass the first value i get from `KVGet` to the `funcValue` func and have these test pass so just want to capture that case.

kv_test.go
cpoile: Ok, I added a test for that, but it's just testing whether the value is passed from line 168 to line 172--not much of a test. However, writing that test made me realize something I would have discovered as soon as I used this in the next PR: that if you pass an `interface{}` into get, it will marshal structs into a map, so `valueFunc` will have an annoying time coercing that back into the original type (map -> marshal -> []byte -> unmarshal -> expected type). I changed it to send `valueFunc` a `[]byte`. `valueFunc` can do what it wants with the []byte (convert it like I've done in the tests, or compare the raw `[]byte`, etc.), and return the `newValue` to be stored.

kv_test.go
cpoile: I'm not sure how that would be possible to do, and not sure how it would be useful...?

kv_test.go
cpoile: > As I think it is possible if you were to change the implementation to always pass the first value i get from KVGet to the funcValue func and have these test pass so just want to capture that case.

Could you clarify this? I don't really understand the problem the test would guard against--if KVGet returned a different value each time it was called, and we give the mocked KVSetWithOptions the wrong expected OldValue, it will fail until we give it the correct OldValue. What would you be testing? Maybe I'm not understanding the example case?

kv.go
```diff
@@ -138,6 +142,51 @@ func (k *KVService) CompareAndDelete(key string, oldValue interface{}) (bool, er
...
+//                         expected to be stored.
+//
+// Returns:
+//
+//       Returns err if the key could not be retrieved (DB error), valueFunc returned an error,
+//               if the key could not be set (DB error), or if the key could not be set (after retries).
+//       Returns nil if the value was set.
+//
+// Minimum server version: 5.18
+func (k *KVService) SetAtomicWithRetries(key string, valueFunc func(oldValue []byte) (newValue interface{}, err error)) error {
```
ali-farooq0, line 166: any reason why `oldValue` is `[]byte` but `newValue` is `interface{}`? should they both be `interface{}`?

ali-farooq0: oh just saw your other comment :D

kv_test.go
ali-farooq0: I would argue what's being passed into the `valueFunc` is quite important for cases where we're mutating the `oldValue` and returning that. If the `oldValue` being passed in is invalid that may result in storing an invalid value.

Regarding passing in `interface{}` vs `[]byte`, is converting the only annoyance there? I wonder if we can write a helper to address that? As it is odd to see `valueFunc` getting `[]byte` but returning `interface{}` and that will lead to confusion. Thoughts?

kv_test.go
ali-farooq0: Could update a flag in here and assert that. But maybe it's not that useful. I was thinking if the client is doing an external operation that has a side effect, calling this func erroneously (in this case) in those scenarios it may lead to problems. But since those cases will be rare maybe it's not worth the hassle. More a correctness concern than anything. 0/5 though if we should assert.

kv_test.go
```diff
@@ -195,6 +197,195 @@ func TestCompareAndDelete(t *testing.T) {
...
+			valueFunc: func(old []byte) (interface{}, error) {
+				type toStore struct {
+					Value int
+				}
+				var fromDB toStore
+				if err := json.Unmarshal(old, &fromDB); err != nil {
+					return nil, err
+				}
+				if fromDB.Value != 1 {
+					return nil, errors.New("old not as expected")
```
ali-farooq0, line 236: would it be better to assert here than returning an error? As i'm afraid that this is relying on the implementation of the method to correctly report this through the API, where in reality it is conceivable that this may not be reported back correctly. Thoughts?

cpoile: > As i'm afraid that this is relying on the implementation of the method to correctly report this through the API, where in reality it is conceivable that this may not be reported back correctly.

I'm not exactly sure what you mean, can you clarify?

kv_test.go
ali-farooq0: Not entirely. Basically what I want to assert is that with every iteration we assert that the value that's being passed into `valueFunc` is different and the return value from `valueFunc` that's used to invoke `KVSetWithOptions` is also different. That would assert the implementation is using the updated values for each iteration. Unless there is a better way to assert that 🤷‍♂️

kv_test.go
cpoile: Okay. No need for a flag, I'm testing the error returned, that will tell you where it failed.

kv_test.go
cpoile: Okay, done.

kv_test.go
ali-farooq0: So here we're returning an error if `fromDB.Value` does not match the value we're expecting, and through that we're asserting that this function should receive the expected `old` value. 

The concern here is that it assumes that this error gets returned through the `SetAtomicWithRetries` API as per its implementation. What if that assumptions was not correct, i.e. what if the implementation is updated to not return that error but rather ignore it, it would then proceed to use that `nil` value in `KVSetWithOptions` and if certain conditions were just right (in the mock setup, etc) this test would pass even though the `oldValue` being passed into this func would not be expected.

My suggestion was to assert directly in this function as opposed to doing it via returning an error. The added benefit of asserting directly in this func is that when it fails you know exactly which condition caused the test to fail, making it easy to debug, rather than figuring out why a certain mock condition wasn't satisfied.

cpoile: First, I don't think it's possible to assert within the call to valueFunc (in a way that would fail the tests). If there is, let me know and I'll do it. It will be quick.

Otherwise, I don't agree that this is a case we need to test for. It's not testing the function, it's testing against every future change that could happen to the code--that's a standard of testing that is never applied to any other PR, so why here? If someone were to change the body of SetAtomicWithRetries so that it ignored the error returned by valueFunc, then it breaks the contract of the function (specified in the comments to the function, and as tested in the tests -- if an error is thrown, it will stop the function and be returned).

Look, I think we're way past the point of high impact here.

ali-farooq0: You can use a function closure to assert, something like:
```
func getValueFunc(t *testing.T) func(old []byte) (interface{}, error) {
	return func(old []byte) (interface{}, error) {
		...
        require.Equal(t, 2, old)
	}
}
```

> It's not testing the function, it's testing against every future change that could happen to the code

Not advocating to test against every future change, that would be impossible :P

Just advocating to lock down the current expected behaviour of the API by asserting using the testing API as opposed to relying on the implementation of the API itself. Especially since this is a plugin API which will be used by many, and therefore has a high impact.

2/5 that we should assert if it's straight forward to do.

kv_test.go
ali-farooq0: @cpoile this pr is good to go once the above concern is addressed:

> Regarding passing in interface{} vs []byte, is converting the only annoyance there? I wonder if we can write a helper to address that? As it is odd to see valueFunc getting []byte but returning interface{} and that will lead to confusion. Thoughts?

cpoile: @ali-farooq0 Yes, I agree it's a little strange to see `valueFunc` take a `[]byte` and return an `interface{}`, but only at first. It makes sense once you realize that the `interface{}` being passed in is not what you expect it to be. You can't cast it to your original type. That is a hidden gotcha that the users have to learn, and they would probably have to relearn it each time. When you pass in a `[]byte` at least it is clear what it is and that you have to interpret it yourself. Good?

kv_test.go
cpoile: Okay, done.

kv_test.go
ali-farooq0: Ahh i see, i see. Do you what the underlying type is that's passed in when it was `interface{}`?

cpoile: By default, structs are marshaled into map[string]interface{}

ali-farooq0: ooo.... gotcha well that's unfortunate :/ your approach makes sense in that case. The other option could be to use `json.RawMessage` but that's essentially the same thing so 0/5. Alright Cool!

PR #50, opened 2020-06-17
Fix to support 5.24

cpoile:
#### OLD:
There's a breaking change on the server 5.24-rc1 for plugins: https://github.com/mattermost/mattermost-server/commit/41e58d976921cfb10a0aefa14a72303aa3c72852 `app/group.go` line 14, 
```
func (a *App) GetGroupByName(name string) (*model.Group, *model.AppError) {
```
is now:
```
func (a *App) GetGroupByName(name string, opts model.GroupSearchOpts) (*model.Group, *model.AppError) {
```
We need to update the `GetGroupByName` call in `plugin-api` when we want to use 5.24-rc1+ (e.g., for autocomplete).

#### Current:
Simply bumping the version number, since the 5.24 change to `post.go` was fixed by larkox in a recent PR.

#### Ticket Link
- no ticket


group.go
```diff
@@ -21,9 +21,9 @@ func (g *GroupService) Get(groupID string) (*model.Group, error) {
 
 // GetByName gets a group by name.
 //
-// Minimum server version: 5.18
-func (g *GroupService) GetByName(name string) (*model.Group, error) {
-	group, appErr := g.api.GetGroupByName(name)
+// Minimum server version: 5.24
+func (g *GroupService) GetByName(name string, opts model.GroupSearchOpts) (*model.Group, error) {
```
ali-farooq0, line 25: ooo this is pretty bad...

https://github.com/mattermost/mattermost-server/commit/41e58d976921cfb10a0aefa14a72303aa3c72852#diff-fafc96d76e739ce68a91df4fc0a3b87fR494 should never been have merged in the first place 😭.

I guess any option of reverting the original change is out the window since 5.24 is already out.

We definitely have consumers of this API (Joram) so it will cause breakage for them. I wonder if we can add the original API back and have another API that allows options? Or maybe turn this one into `...Options`.

cpoile: Ah, yes, that's better. I felt this was pretty wrong.  one sec...

cpoile: Yah, I understand, it's unfortunate the api change was merged. But I guess it can't be helped. @lieut-data ?

hanzei: I can't understand how this change got into 5.24. Given that a dot release might be cut (https://community.mattermost.com/core/pl/ti8s6jfmujg9uxsnrgw96farth), there is room for reverting the change. Thoughts @lieut-data?

lieut-data: There's another discussion over at https://community.mattermost.com/core/pl/6xdimkjugt819fkppuqbnddzwe, but yes, I think that's the general consensus.

cpoile:
Waiting on 5.24.1, then will update this PR to only include the changes that `Post`'s `RWMutex` introduced.

cpoile:
@lieut-data I did some testing, and it looks like everything works fine as long as you depend on 5.24.1 and the latest commit from `plugin-api` -- so this version bump isn't needed anymore. Should I close it then?

lieut-data:
Sounds good!

PR #51, opened 2020-06-18
Add OAuther logic

larkox:
#### Summary
Add a simple interface to implement OAuth in any plugin.

#### Ticket Link
None


experimental/oauther/oauth2.go
```diff
@@ -0,0 +1,101 @@
...
+	"encoding/json"
+
+	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-plugin-api/experimental/bot/logger"
+	"github.com/mattermost/mattermost-server/v5/plugin"
+	"golang.org/x/oauth2"
+)
+
+const (
+	oAuth2StateTimeToLive = 300 // seconds
```
lieut-data, line 16: Should this be configurable?

experimental/oauther/oauth2.go
```diff
@@ -0,0 +1,101 @@
...
+	oAuthURL        string
+	connectedString string
+	config          *oauth2.Config
+	logger          logger.Logger
+}
+
+func NewOAuther(
+	r *mux.Router,
+	api plugin.API,
+	pluginURL,
```
lieut-data, line 39: This seems like something the component can figure out for itself, vs. needing to be set.

experimental/oauther/oauth2.go
```diff
@@ -0,0 +1,101 @@
...
+	oAuthURL,
```
lieut-data, line 40: Would it make sense to have a sane default for this instead of requiring it to be configurable?

experimental/oauther/oauth2.go
```diff
@@ -0,0 +1,101 @@
...
+		pluginURL:       pluginURL,
+		oAuthURL:        oAuthURL,
+		config:          oAuthConfig,
+		connectedString: connectedString,
+		logger:          loggerBot,
+	}
+
+	o.config.RedirectURL = pluginURL + oAuthURL + "/complete"
+
+	oauth2Router := r.PathPrefix(oAuthURL).Subrouter()
```
lieut-data, line 60: Instead of accepting a router and binding the client to `gorilla/mux`, can we make oAuther an `http.Handler` instead?

experimental/oauther/oauth2.go
```diff
@@ -0,0 +1,101 @@
...
+
+	return o
+}
+
+func (o *oAuther) GetURL() string {
+	return o.pluginURL + o.oAuthURL + "/connect"
+}
+
+func (o *oAuther) GetToken(userID string) (*oauth2.Token, error) {
+	rawToken, appErr := o.api.KVGet(o.getTokenKey(userID))
```
lieut-data, line 72: Would this be an opportunity to accept the plugin-api-wrapper client instead, so we don't have to re-implement JSON serialization?

experimental/oauther/oauth2.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (o *oAuther) getTokenKey(userID string) string {
+	return o.storePrefix + "token_" + userID
+}
+
+func (o *oAuther) getStateKey(userID string) string {
+	return o.storePrefix + "state_" + userID
+}
+
+func (o *oAuther) Deauth(userID string) error {
```
lieut-data, line 94: I've never been a fan of unnecessary abbreviations. Is `Deauth` an official term in oAuth? Is there a more appropriate term we could use?

experimental/oauther/oauth2_complete.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+func (o *oAuther) oauth2Complete(w http.ResponseWriter, r *http.Request) {
+	authedUserID := r.Header.Get("Mattermost-User-ID")
```
lieut-data, line 15: As an aside, I'd love to be able to reuse `http.MattermostAuthorizationRequired` as exposed by this package. (Our `http` package in this repository might also be the long-term home for this code.)

experimental/oauther/oauth2_complete.go
```diff
@@ -0,0 +1,89 @@
...
+	storedState, appErr := o.api.KVGet(o.getStateKey(authedUserID))
+	if appErr != nil {
+		o.logger.Warnf("oauth2Complete: cannot get state, err=%s", appErr.Error())
+		http.Error(w, "cannot get stored state", http.StatusInternalServerError)
+		return
+	}
+
+	if string(storedState) != state {
+		o.logger.Debugf("oauth2Complete: state mismatch")
+		http.Error(w, "state does not mach", http.StatusUnauthorized)
```
lieut-data, line 38:
```suggestion
 		http.Error(w, "state does not match", http.StatusUnauthorized) 
```

experimental/oauther/oauth2_complete.go
```diff
@@ -0,0 +1,89 @@
...
+		return
+	}
+
+	userID := strings.Split(state, "_")[1]
+	if userID != authedUserID {
+		o.logger.Debugf("oauth2Complete: authed user mismatch")
+		http.Error(w, "authed user is not the same as state user", http.StatusUnauthorized)
```
lieut-data, line 45: This is effectively an information leak: we've told the client something they didn't need to know. Can we generalize these error so it's impossible to distinguish the cases (fine to keep logging the particulars).

experimental/oauther/oauth2_complete.go
```diff
@@ -0,0 +1,89 @@
...
+		return
+	}
+
+	ctx := context.Background()
+	tok, err := o.config.Exchange(ctx, code)
```
lieut-data, line 50: Can we avoid unnecessary abbreviations? Should this be `token`?

experimental/oauther/oauth2_complete.go
```diff
@@ -0,0 +1,89 @@
...
+	rawToken, err := json.Marshal(tok)
+	if err != nil {
+		o.logger.Errorf("oauth2Complete: could not marshal token, err=%s", err.Error())
+		http.Error(w, "cannot marshal the token", http.StatusInternalServerError)
+		return
+	}
+	appErr = o.api.KVSet(o.getTokenKey(userID), rawToken)
+	if appErr != nil {
+		o.logger.Errorf("oauth2Complete: cannot store the token, err=%s", appErr.Error())
+		http.Error(w, "cannot store token", http.StatusInternalServerError)
```
lieut-data, line 66: Missing a `return` here.

experimental/oauther/oauth2.go
larkox: We could use a function like:
`NewOAuther(..., options [string]interface{})`
If the option is not there, just add the default. Nevertheless, I am not specially a fan of the options maps (hard to guess what options you can pass).
Also, I prefer to require this configuration, to avoid clashes. If I set the oAuthURL, I know I cannot do anything with those URLs. The same for KVStore prefix. If some errors start to pop up because of the default values, it is harder to track the problem. Better safe than sorry, I think.

experimental/oauther/oauth2.go
larkox: I would like to get rid of gorilla/mux and add something more abstract, so gorilla/mux or any other service could be used for this. Nevertheless, I was not able to think about a better way to do this, and gorilla/mux is broadly used among our plugins.

experimental/oauther/oauth2.go
larkox: If we pass the whole plugin.API, yes, it is something it can figure out itself. Nevertheless, I think that passing the whole api is not the correct approach, specially since we only want the API to access the KVStore. I changed this to pass a KVStore interface (that should be met by the StoreService) and remove the plugin.API reference.

experimental/oauther/oauth2.go
larkox: This is partly a magic number. 5 minutes feel more than enough to go through an OAuth process, and short enough to avoid being used in a wrong manner. I do not see many cases where this number must be larger or smaller, so adding some configuration for this would just mean more complex code and initialization without real usage.

experimental/oauther/oauth2_complete.go
larkox: That sounds like a great idea, but maybe we should put those in a http package and implement them on a different PR.

experimental/oauther/oauth2_complete.go
```diff
@@ -21,49 +20,50 @@ func (o *oAuther) oauth2Complete(w http.ResponseWriter, r *http.Request) {
...
-		http.Error(w, "cannot marshal the token", http.StatusInternalServerError)
+		o.logger.Errorf("oauth2Complete: cannot store the token, err=%s", err.Error())
+		http.Error(w, "Internal server error", http.StatusInternalServerError)
 		return
 	}
-	appErr = o.api.KVSet(o.getTokenKey(userID), rawToken)
-	if appErr != nil {
-		o.logger.Errorf("oauth2Complete: cannot store the token, err=%s", appErr.Error())
-		http.Error(w, "cannot store token", http.StatusInternalServerError)
+	if !ok {
```
lieut-data, line 63: This is an interesting problem -- the `Set` call will never return `err == nil` and `!ok` unless the `SetAtomic` configuration is set. I don't like that the signature forces clients like yours to have to check the `ok` case when it doesn't matter, since the signature looks like it might matter. Fine for now, but just something to consider as we iterate on the plugin API wrapper.

experimental/oauther/oauth2.go
```diff
@@ -35,17 +36,18 @@ type oAuther struct {
...
-	api plugin.API,
+	store common.KVStore,
 	pluginURL,
 	oAuthURL,
 	storePrefix,
 	connectedString string,
 	onConnect func(userID string, token *oauth2.Token),
 	oAuthConfig *oauth2.Config,
 	loggerBot logger.Logger,
+	test http.Handler,
```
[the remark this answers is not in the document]
larkox: My bad. Added it to test some things, and forgot to remove it.

experimental/oauther/oauth2.go
levb: @lieut-data functional options to the rescue? :)

experimental/oauther/oauth2.go
levb: 5/5 that I personally dislike gorilla/mux, specifically for its lack of support for request-time "variables" (context) - they resort to the go context which is the wrong thing to do for the use-case.

Moving the dependency to the caller by providing an http.Handler sounds like a good approach.

experimental/oauther/oauth2.go
larkox: I have made a take on functional options. If they look nice, I may start using them on other libraries.

experimental/oauther/oauth2.go
larkox: I also did this, but not 100% sure I did what was expected. Any feedback on these changes would be welcomed.

levb:
@larkox @lieut-data what's up with lint settings here? (1/5 should fix even if just disable)

larkox:
@levb This was in part my mistake. The panel and flow PR was merged without checking the stile, and the lint was only working on master, not PRs. Therefore, most of these lint problems come from master. #52 is solving those, so once it is merged, most if not all the lint errors will be fixed.

experimental/oauther/oauth2_complete.go
```diff
@@ -0,0 +1,93 @@
...
+				</script>
+			</head>
+			<body>
+				<p>%s</p>
+			</body>
+		</html>
+		`, o.connectedString)
+
+	w.Header().Set("Content-Type", "text/html")
+	_, _ = w.Write([]byte(html))
```
hanzei, line 88: Very, very nit pick: Logging the error might help debugging some rare cases

experimental/oauther/oauth2.go
```diff
@@ -0,0 +1,138 @@
...
+		oAuthURL:              DefaultOAuthURL,
+		connectedString:       DefaultConnectedString,
+		oAuth2StateTimeToLive: DefaultOAuth2StateTimeToLive,
+	}
+
+	for _, option := range options {
+		option(o)
+	}
+
+	o.config.RedirectURL = o.pluginURL + o.oAuthURL + "/complete"
```
hanzei, line 94: E.g. GitHub doesn't allow OAuth Applications to set the `RedirectURL` itself. Admins have to set it via a UI. Is this the same for other OAuth applications?

larkox: As far as I know, you set the RedirectURL through the UI, and the configuration must match the RedirectURL. If the configuration and what is set on the UI mismatch, you will receive an error message during the OAuth flow saying that the redirect URL does not match.

Is anyhow different in Github?

experimental/oauther/oauth2.go
```diff
@@ -0,0 +1,138 @@
...
+
+- store: A KVStore to store the data of the OAuther.
+
+- l Logger: A logger to log errors during authorization.
+
+- options: Optional options for the OAuther. Available options are StorePrefix, OAuthURL, ConnectedString and OAuth2StateTimeToLive.
+*/
+func NewOAuther(
+	pluginURL string,
+	oAuthConfig *oauth2.Config,
```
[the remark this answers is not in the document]
larkox: Will remove the pointer.

experimental/common/kvstore.go
```diff
@@ -2,6 +2,79 @@ package common
...
+
+type KVStore interface {
+	// Set stores a key-value pair, unique per plugin.
+	// Keys prefixed with `mmi_` are reserved for use by this package and will fail to be set.
+	//
+	// Returns (false, err) if DB error occurred
+	// Returns (false, nil) if the value was not set
+	// Returns (true, nil) if the value was set
+	//
+	// Minimum server version: 5.18
```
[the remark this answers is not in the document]
larkox: No special reason. I can remove it.

experimental/oauther/oauth2.go
hanzei: AFAIK GitHub only uses the URL from the UI. 

The reasoning I'm bringing this question up is that it's weird to require a `pluginURL` from a library that should help you setup an URL. Any ideas on how to drop that parameter?

larkox: It is complicated to drop it. We need the plugin URL to also get the connect URL. We could require the whole API, and look for the URL ourselves, but I prefer to ask for "just what is needed".

Regarding the RedirectURL, I wonder if passing it to Github will pose any problem. If it does, then we would have to add some option to remove or to add it.

lieut-data: Could we add a second constructor, `NewFromClient(pluginapi *pluginapi.Client, ...)` that streamlines this?

larkox: In my opinion, using the pluginapi client have not solved many pain points, but I added it.

Right now, you can omit the store (you can take it from the plugin api) and you don't have to set the URL, but it still needs the plugin id (AFAIK it is only on the manifest, and pluginapi does not have access to the manifest).

We could consider creating also a default logger, but since there is more configuration to be handled on the logger, I think it would be best to leave the logger as it is.

experimental/oauther/oauth2.go
```diff
@@ -0,0 +1,138 @@
...
+	onConnect func(userID string, token *oauth2.Token),
```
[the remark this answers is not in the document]
hanzei: Did you intent to change this into 
```suggestion
	onConnect func(userID string, token oauth2.Token),
```
?

experimental/oauther/oauth2.go
```diff
@@ -0,0 +1,173 @@
...
+
+- onConnect: What to do when the Authorization process is complete.
+
+- l Logger: A logger to log errors during authorization.
+
+- options: Optional options for the OAuther. Available options are StorePrefix, OAuthURL, ConnectedString and OAuth2StateTimeToLive.
+*/
+func NewFromClient(
+	client *pluginapi.Client,
+	pluginID string,
```
hanzei, line 118: Actually, there is a `GetManifest` method in `pluginapi.Client`. Hence `pluginID` isn't needed.

larkox: Not sure how I did not find it. Thanks!

experimental/oauther/oauth2.go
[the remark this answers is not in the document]
larkox: I had the intention, but not sure how I forgot that one.

PR #52, opened 2020-06-18
Fix lint

larkox:
#### Summary
Fix lint errors introduced by #47 

#### Ticket Link
None


post.go
```diff
@@ -19,6 +19,7 @@ func (p *PostService) CreatePost(post *model.Post) error {
 		return normalizeAppErr(appErr)
 	}
 
+	//nolint:govet //lock value is not relevant to consumer
```
larkox, line 22: @lieut-data I wanted to have your attention on these two nolints. They are not introduced by my code, and I wanted to double check that the reasoning is in place.
The problem that the linter drops is:
> copylocks: assignment copies lock value to *post: github.com/mattermost/mattermost-server/v5/model.Post contains sync.RWMutex

experimental/freetextfetcher/store.go
```diff
@@ -48,9 +48,9 @@ func (fts *freetextStore) StopFetching(userID string) error {
 	return fts.client.KV.Delete(fts.getKey(userID))
 }
 
-func (fts *freetextStore) ShouldProcessFreetext(userID, fetcherID string) (bool, string, error) {
+func (fts *freetextStore) ShouldProcessFreetext(userID, fetcherID string) (shouldProcess bool, payload string, err error) {
```
lieut-data, line 51: Curious what linting rule triggered this change? Named return values are usually exceptional.

post.go
lieut-data: I think this was from an upstream change in mattermost-server -- I think we can handle this by using the new `Clone()` method on `model.Post`.

experimental/freetextfetcher/store.go
hanzei: https://go-critic.github.io/overview#unnamedresult is the rule in question. Maybe enabling `opinionated` wasn't a good idea: https://github.com/mattermost/mattermost-plugin-api/blob/d4df2d1e8f88e7d892eac4f7aac83c87dd22dde8/.golangci.yml#L47

PR #53, opened 2020-06-18
Add telemetry

larkox:
#### Summary
Add telemetry functionality

#### Ticket Link
None


experimental/telemetry/rudder.go
```diff
@@ -0,0 +1,46 @@
+package telemetry
+
+import rudder "github.com/rudderlabs/analytics-go"
+
+const rudderDataPlaneURL = "https://pdat.matterlytics.com"
+
+var rudderWriteKey string
```
lieut-data, line 7: @larkox, can you add some GoDoc to these files to help clarify intent and usage? I have a bunch of comments, but don't want to make assumptions without better understanding how things are meant to be used.

experimental/telemetry/tracker.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+func (t *tracker) Track(event string, properties map[string]interface{}) {
+	if !t.enabled || t.client == nil {
+		return
+	}
+
+	event = t.telemetryShortName + "_" + event
+	properties["PluginID"] = t.pluginID
```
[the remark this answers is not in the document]
larkox: Not 100% sure about the reasoning, but the people at analytics preferred one single write key for all plugins.

experimental/telemetry/tracker.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Tracker defines a telemetry tracker
+type Tracker interface {
+	//Track registers an event throught the configured telemetry client
+	Track(event string, properties map[string]interface{})
+	//TrackUserEvent registers an event through the configured telemetry client associated to a user
+	TrackUserEvent(event string, userID string, properties map[string]interface{})
+}
+
+// Client defines a telemetry client
```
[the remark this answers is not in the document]
larkox: I like to use interfaces in order to allow future development more easy (or even "client side" customizations). If we were to limit this Rudder, the code will be simpler, but would discourage the development of new clients.
Nevertheless, I am 0/5 on which path to take on this.

experimental/telemetry/rudder.go
```diff
@@ -0,0 +1,56 @@
...
+
+	return &rudderWrapper{client: client}, nil
+}
+
+type rudderWrapper struct {
+	client rudder.Client
+}
+
+func (r *rudderWrapper) Enqueue(t Track) error {
+	err := r.client.Enqueue(rudder.Track{
```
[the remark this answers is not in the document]
larkox: We probably can, but I just get a bit obsessed with the "not nil interface pointing to nil" problem we get sometimes with AppError in other parts of the code.
Will change.

experimental/telemetry/rudder.go
```diff
@@ -0,0 +1,56 @@
+package telemetry
+
+import rudder "github.com/rudderlabs/analytics-go"
+
+// rudderDataPlaneURL is common for all Mattermost Projects
+const rudderDataPlaneURL = "https://pdat.matterlytics.com"
```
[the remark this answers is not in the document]
larkox: The repo and the package is not intended just for Mattermost-authored plugins, but we will be the main consumer of the package. Therefore, it makes sense to make our lives easier. We could change the name of the variable to "defaultRudderDataPlaneURL" or "mattermostRudderDataPlaneURL" to make it more clear.
Nevertheless, any user that want to use their own service instead of mattermost's one can always use the "NewRudderClientWithCredentials".

experimental/telemetry/rudder.go
```diff
@@ -0,0 +1,56 @@
...
+
+// rudderWriteKey is set during build time adding the following line to build/custom.mk
+// LDFLAGS += -X "github.com/mattermost/mattermost-plugin-api/experimental/telemetry.rudderWriteKey=$(MM_RUDDER_WRITE_KEY)"
+// MM_RUDDER_WRITE_KEY environment variable must be set also during CI to "1dP7Oi78p0PK1brYLsfslgnbD1I"
```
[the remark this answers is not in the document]
larkox: The idea here is to use the same write key along all plugins (incident response is a special case, so it makes sense it has its own key). The idea is to make our lives easier, and anyone that want to use its own rudder instance can just change the values.
I could change the phrasing to something like
> ... must be set also during CI to your production write key ("..." for Mattermost projects)

larkox:
@lieut-data Thanks for your comments. This code is now making its way (mostly copy pasted) into Jira and Microsoft Calendar. Nevertheless, in order to have the code "battle tested", we need it "ready for battle" and somewhere we can just use it. Every time we we copy paste the code from one plugin to another, we may end up making changes that pollute the code with the implementation's internals. I have seen this in various of the extractions I have done on MSCalendar, and would like to avoid that.
Once we have this merged, we will start to implement this feature in the rest of our plugins, and really battle test the interface and use case.

Regarding the location of these packages, I have always been 0/5 it should go on plugin-api repo. But a "mattermost-plugin-utils" repo had also many detractors. From my side, I have no strong opinion on where should these packages reside, as long as they are shareable among the plugins.

experimental/telemetry/rudder.go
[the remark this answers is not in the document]
larkox: We can also set it up as a variable, so anyone can change that too the same way as rudderWriteKey, but 0/5 on that.

experimental/telemetry/tracker.go
```diff
@@ -0,0 +1,94 @@
...
+type Client interface {
+	// Enqueue adds a tracker event (Track) to be registered
+	Enqueue(t Track) error
+	// Close closes the client connection, flushing any event left on the queue
+	Close() error
+}
+
+// Track defines an event ready for the client to process
+type Track struct {
```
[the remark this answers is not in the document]
larkox: After a second thought, it makes sense this is exported. It is one of the incoming values of the Client interface. If we were to remove the client interface and use directly rudder, we won't need it at all, but since I would prefer to keep the interface to allow customization and extension of the package, this must be exported so people can handle it in their implementations.

levb:
>  I'd be inclined to revisit whether this should live in a separate repository

If it makes it easier for us to iterate and evolve - I'd support having a separate repo. 2/5 the primary concern, as @larkox stated, is to enable code sharing between our plugin and mature them faster. Assuring the API maturity right away for public consumption - not a high priority until "field-proven", as a shared API.

A name? `mattermost-plugin-utils` - we'd be back at square 1 (acceptable to me); `mattermost-plugin-experimental` seems weird to use in production code.

experimental/telemetry/tracker.go
```diff
@@ -0,0 +1,94 @@
...
+		logger:             l,
+	}
+}
+
+func (t *tracker) Track(event string, properties map[string]interface{}) {
+	if !t.enabled || t.client == nil {
+		return
+	}
+
+	event = t.telemetryShortName + "_" + event
```
agarciamontoro, line 75: If `telemetryShortName` is empty, this will look weird. We can either check the string in `NewTracker` and throw an error if it is empty or check it here, and add the `pluginID` instead, for example.

experimental/telemetry/tracker.go
```diff
@@ -0,0 +1,94 @@
...
+	properties["PluginID"] = t.pluginID
+	properties["PluginVersion"] = t.pluginVersion
+	properties["ServerVersion"] = t.serverVersion
+
+	err := t.client.Enqueue(Track{
+		UserID:     t.diagnosticID,
```
agarciamontoro, line 81: Can we add a comment here about why UserID is not the id of a user? :joy:

experimental/telemetry/tracker.go
```diff
@@ -0,0 +1,94 @@
...
+		Properties: properties,
+	})
+
+	if err != nil {
+		t.logger.Warnf("cannot enqueue telemetry event, err=%s", err.Error())
+	}
+}
+
+func (t *tracker) TrackUserEvent(event, userID string, properties map[string]interface{}) {
+	properties["UserID"] = userID
```
agarciamontoro, line 92: Doesn't this clash with the UserID that we set to diagnosticID?

experimental/telemetry/doc.go
```diff
@@ -0,0 +1,16 @@
...
+// to allow having different keys for production and development.
+// If you are working on a Mattermost project, the data plane URL is already set.
+// In order to add the write key for both development and production environment,
+// add the following lines to build/custom.mk
+//   ifndef MM_RUDDER_WRITE_KEY
+//     MM_RUDDER_WRITE_KEY = 1d5bMvdrfWClLxgK1FvV3s4U1tg
+//   endif
+//   LDFLAGS += -X "github.com/mattermost/mattermost-plugin-api/experimental/telemetry.rudderWriteKey=$(MM_RUDDER_WRITE_KEY)"
+// MM_RUDDER_WRITE_KEY environment variable must be set also during CI
+// to the production write key ("1dP7Oi78p0PK1brYLsfslgnbD1I").
```
agarciamontoro, line 12: I needed to read this twice to understand why there were two different sample keys here. Could we specify that the first is for development?

experimental/telemetry/doc.go
```diff
@@ -0,0 +1,16 @@
...
+// If you want to use your own data plane URL, add also this line and
+// make sure the MM_RUDDER_DATA_PLANE_URL variable is set.
```
agarciamontoro, line 14: Nit:
```suggestion
// make sure the MM_RUDDER_DATA_PLANE_URL environment variable is set.
```

experimental/telemetry/tracker.go
```diff
@@ -0,0 +1,94 @@
+package telemetry
+
+import "github.com/mattermost/mattermost-plugin-api/experimental/bot/logger"
+
+// Tracker defines a telemetry tracker
+type Tracker interface {
+	// Track registers an event through the configured telemetry client
+	Track(event string, properties map[string]interface{})
```
ali-farooq0, line 8: nit: apologies for the drive-by review, wondering if this should be `TrackEvent` for consistency with `TrackUserEvent`?

experimental/telemetry/tracker.go
```diff
@@ -0,0 +1,94 @@
...
+// - pluginVersion
+// - telemetryShortName: Short name for the plugin to use in telemetry. Used to avoid dot separated names like com.company.pluginName
+// - enableDiagnostics: Whether the system has enabled sending telemetry data. If false, the tracker will not track any event.
+// - l Logger: A logger to log any error related with the telemetry tracking.
+func NewTracker(
+	c Client,
+	diagnosticID,
+	serverVersion,
+	pluginID,
+	pluginVersion,
```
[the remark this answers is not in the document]
larkox: I am 0/5 on this. If we do not use the struct, we have something like:
```
New(a,b,c,d,e)
```
If we use the struct, we have something like:
```
New(a, config{b:b, c:c, d:d, e:e})
```
Also, if the API was to change for any reason adding more parameters, the implementation after a version bump will not fail, and start using the 0 value of the new configuration, which might not been desirable. Having it as a parameter will make the code fail on compile time, making you aware that there has been a change there.

experimental/telemetry/tracker.go
larkox: This one is set on the properties. The other one is set on the UserID field.

larkox:
@agarciamontoro regarding coupling more the telemetry to the config and not relying on a proper implementation, I like the idea by itself, but I do not see how to gracefully handle it as a separate package.
The best idea I have at the moment is to expose an `OnConfigurationChange` that accept a `model.Config`, and ask the implementation to call that on each OnConfigurationChange (something similar to what you are doing with the listener over there).
Nevertheless, some other things may change due to configuration that we cannot control (mainly looking at the client and the logger), so probably creating the object from 0 is better.

experimental/telemetry/tracker.go
agarciamontoro: Isn't it the same? Let me take a look

agarciamontoro: Ok, it looks like this would get sent as 
```json
{
  "userId": "diagnostic id",
  "properties": {
    "UserID": "..."
  }
}
```

Looks good, sorry for the confusion!

experimental/telemetry/tracker.go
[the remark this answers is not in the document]
agarciamontoro: That's a good argument, let's stick with your solution.

agarciamontoro:
@larkox I agree that it will require some kind of management from the user either way. Let's go with this and see if we can iterate further in the future.

experimental/telemetry/tracker.go
```diff
@@ -0,0 +1,98 @@
+package telemetry
+
+import "github.com/mattermost/mattermost-plugin-api/experimental/bot/logger"
+
+// Tracker defines a telemetry tracker
+type Tracker interface {
+	// Track registers an event through the configured telemetry client
```
ali-farooq0, line 7:
```suggestion
	// TrackEvent registers an event through the configured telemetry client
```

larkox:
@DHaussermann I will skip QA Review for this, since it should be tested while integrated on a plugin.

PR #57, opened 2020-07-03
Logger improvements

larkox:
#### Summary
- Use functional options
- Use pluginapi compatible interface
- Add documentation
- Add the possibility to track the logs on a telemetry client

#### Ticket Link
None


experimental/bot/logger/config.go
```diff
@@ -3,19 +3,25 @@
...
+// Config contains all configuration needed by the logger.
 type Config struct {
 	// AdminLogLevel is "debug", "info", "warn", or "error".
 	AdminLogLevel string
 
 	// AdminLogVerbose: set to include full context with admin log messages.
 	AdminLogVerbose bool
+
+	// TelemetryLogLevel is "debug", "info", "warn", or "error".
+	TelemetryLogLevel string
```
levb, line 15: 1/5 prefer an wrapper `NewTelemetryLogger(Logger)Logger` to options and bundling everything in 1 type. TelemtryLogger could be completely separated, no?

levb: On that note, the `MessageToAdmin` business could be implemented as a `CcLogger` with its own way of initialization (a list of teams - channels/users)

hanzei: If I understand your proposal correctly, the code would now works like this

```go
log := logger.New(api, logger.LogToTelemetry("debug", tracker))
```
vs. 
```go
log := logger.NewTelemetryLogger("debug", tracker, logger.New(api))
```
is what you are proposing.

I like the functions options more, as it makes clear that a logger can be an admin and telemetry logger.

experimental/bot/logger/config.go
```diff
@@ -3,19 +3,25 @@
...
 
+// Config contains all configuration needed by the logger.
 type Config struct {
 	// AdminLogLevel is "debug", "info", "warn", or "error".
 	AdminLogLevel string
 
 	// AdminLogVerbose: set to include full context with admin log messages.
 	AdminLogVerbose bool
+
+	// TelemetryLogLevel is "debug", "info", "warn", or "error".
```
hanzei, line 14: I would like to see these variables getting `type`d

experimental/bot/logger/config.go
larkox: If I understand correctly, the end result would be something like:
```
func NewTelemetryLogger(logger Logger, logLevel string, tracker telemetry.Tracker) Logger {
   return logger.New(logger, LogToTelemetry(logLevel, tracker)
}
```

The problem with that is that Logger is a interface, so I cannot affect the inners of the object. I could go with a sealed interface, but would make the code much more complicated. The other option I see is instead of "promoting a logger" having a utility function to make faster what you want.
```
func NewTelemetryLogger(api common.LogAPI, logLevel string, tracker telemetry.Tracker) Logger {
   return logger.New(api, LogToTelemetry(logLevel, tracker))
}
```

larkox: There is another option to "promote the logger", adding a WithTracker or WithAdminCC function to the Interface. That way we could do either:
```
l := New(client)
l.WithTracker(LogLevelDebug, tracker)
```
or
```
l := New(client)
l = NewTelemetryLogger(l, LogLevelDebug, tracker)
```
and NewTelemetryLogger would be defined as:
```
func NewTelemetryLogger(logger Logger, logLevel string, tracker telemetry.Tracker) Logger {
   return logger.WithTracker(logLevel, tracker)
}
```

larkox: After discussing offline with @levb , I have made some changes to decouple the default logger from the telemetry and admincc loggers. @hanzei , I re-request your review since the approach is quite different, in case you want to give more comments.

experimental/bot/logger/default_logger.go
```diff
@@ -25,6 +25,15 @@ New creates a new logger.
...
 - options: Optional parameters for the Logger. Available options are MessageToAdmins and LogToTelemetry.
+
+Example:
+  client := pluginapi.NewClient(p.api)
+  dmer := poster.New(...)
+  tracker := telemetry.NewTracker(...)
+  logger := New(client,
+	MessageToAdmins(LogLevelWarn, false, dmer, "000000000000000", "0000000010000"),
+	LogToTelemetry(LogLevelDebug, tracker),
+  )
```
hanzei: Godoc supports documenting examples directly: https://golang.org/pkg/testing/#hdr-Examples. It's the idiomatic way to write examples.

experimental/bot/logger/admincclogger/admincc_logger.go
```diff
@@ -0,0 +1,92 @@
...
+ids provided through the dmer provided, about all events below the logLevel. If logVerbose is set,
+it will also send the context.
+
+- l Logger: A logger to promote.
+
+- dmer DMer: A DMer to send the messages to the admins.
+
+- logLevel: The highest type of message to be stored in telemetry.
+
+- logVerbose: Whether the log context should be messaged to the admins.
```
hanzei, line 30: `logVerbose` isn't a self explaining name. Maybe `logContext` or `includeContext` instead?

experimental/bot/logger/telemetrylogger/telemetry_logger.go
```diff
@@ -0,0 +1,74 @@
...
+	tracker  telemetry.Tracker
+}
+
+/*
+New promotes the provided logger into a telemetry logger, storing all events below the logLevel
+through the tracker.
+
+- l Logger: A logger to promote.
+
+- logLevel: The highest type of message to be stored in telemetry.
```
hanzei, line 22: What is the use case of a logeLevel? I can't think of one.

experimental/bot/logger/telemetrylogger/telemetry_logger.go
```diff
@@ -0,0 +1,74 @@
...
+
+- tracker: The telemetry tracker to store the messages.
+*/
+func New(l logger.Logger, logLevel logger.LogLevel, tracker telemetry.Tracker) logger.Logger {
```
hanzei, line 26: It's idiomatic in go to return actual implementations instead of interfaces. What is the reasoning behind not doing this here? (`telemetryLogger` should be exported then)

hanzei, line 26: What is the value of passing a `Logger` here instead of a `common.LogAPI`? Why do I have to create a `Logger` first?

experimental/bot/logger/telemetrylogger/telemetry_logger.go
larkox: We may want to send to telemetry only Error level logs, or anything that is not a Debug log. That is the reason to be of logLevel.

experimental/bot/logger/telemetrylogger/telemetry_logger.go
larkox: This way, if we want to create a telemetry logger that also send copies to the admins, we can compose like layers. Something like this:
```
l := admincclogger.New(telemetrylogger.New(logger.New(...), ...), ...)
```
We could also compose something more complex like:
```
l := admincclogger.New(admincclogger.New(logger.New(...), LogLevelError, sysadmins), LogLevelInfo, pluginAdmins)
```

larkox: One upside of returning the interface is that we are sure on compile time that we are conformant with the interface. Nevertheless, there are other ways to achieve the same.

The main reason is personal preference. I prefer to keep things as "standard" as possible so they can be generalized. If we start returning the struct, we might end up declaring directly the struct, or modifying things that we understand invariant on the struct and making it more prone to errors. If we just interact with the interface, the usage is more stiff, but it is harder to "make mistakes".

Also, I would not like other packages working over "telemetryLogger" objects, since will make things way harder to mock and generalize.

Nevertheless, as I said, it is a personal preference. I am open to discussion or changes.

larkox: We can add another function NewFromAPI to simplify this, but not sure if needed.

experimental/bot/logger/default_logger.go
larkox: Noted. Nevertheless, it seems like there is no need for examples now. What do you think?

experimental/bot/logger/telemetrylogger/telemetry_logger.go
levb: I've seen strong opinions expressed on the subject. I do not have a personal strong preference, but lean towards the public `NewXXX` methods returning (public) interfaces. As practical, `newXXX` returning concrete types can be added for intra-package use.

experimental/bot/logger/default_logger.go
hanzei: This is defiantly a non-blocking suggestion. Still, I consider every example helpful and valuable.

experimental/bot/logger/telemetrylogger/telemetry_logger.go
hanzei: I still don't quite get the use case. When is the decision to send on a certain level made? On build time? Is there a case where it's decided on run time?

experimental/bot/logger/telemetrylogger/telemetry_logger.go
hanzei: I don't have a strong option either. Keeping things as generalized as possible is a good point :+1:

experimental/bot/logger/telemetrylogger/telemetry_logger.go
larkox: In this case, for telemetry logger, I cannot think any "real scenario" where deciding on run time make sense. For Admin logger, we may want to have a setting somewhere, and change the log level based on the setting.
But on the weird scenario that the plugin developer uses this telemetry logger to log into its own logging system, and has a setting to decide which logLevel to log, that would be a use case for this to be set on run-time.

hanzei: I'm 1/5 that we should go with the simpler approach here and drop the logLevel for the telemetry logger. And if we come up with a real world use case, we can still add it later.

larkox: But what about the case we want to decide on build time? Like "log to telemetry only error messages"?

hanzei: If the decision is made on build time there are already tools like to build flags to handle that case.

larkox: I will keep the logLevel for versatility, even if it makes since slightly more complex. Nevertheless, I keep you comments in mind.

PR #58, opened 2020-07-06
Add o auther payload

larkox:
#### Summary
Sometimes we want to pass more information into the OAuth process. Therefore, I added a payload.
Many times we pass more information into the state, but it felt that it may become too crowded, so I decided to store it in the KVStore. I am open to any suggestions and improvements over this.

#### Ticket Link
None


levb:
@DHaussermann skipping QA here, FYI

PR #59, opened 2020-07-06
Add tests to Poster

larkox:
#### Summary
Add tests to Poster implementation.

#### Ticket Link
None


larkox:
I added some tests, but I do not like having to pass two different apis (Post and Channel) in the creation of the object. Passing the client and mocking the plugin api will mean knowing the implementation of pluginapi, which I prefer the test to ignore. Any ideas? Or is this the path to follow?

larkox:
Another solution could be delegate the DM function down to plugin API, but I am not sure if that is a desirable approach.

lieut-data:
@larkox, agree that our mocking strategy leaves something to be desired. I'm really glad you're driving some of these questions, since I think we'll land somewhere awesome after we keep experimenting.

I'm intrigued by your idea of promoting `DM` to the plugin api wrapper. It was definitely the intent for the wrapper to contain "helper" functionality, including the deprecated "helpers" from our existing `mattermost-server/v5/plugin` package. Would you want to prototype that change for review?

larkox:
All test added, and changed a bit the way it behaves after merging the DM into the postAPI.

larkox:
Any simplification of CreatePost, Ephemeral, or UpdatePost is a good candidate for promotion, since we usually use them "the simple way". But I am not sure whether it makes so much sense to do it without a poster, since it is mainly a mapping of the arguments into the object to pass.
```
// Without poster
func CreatePost(userID, channelID, message) {
  return api.CreatePost(Post{
    userID: userID,
    channelID: channelID,
    message: message
  })
}
```

```
// with poster
func (p) CreatePost(channelID, message string) {
  p.api.CreatePost(Post{
    UserID: p.id,
    channelID: channelID,
    message: message,
  })
}
```

larkox:
After a discussion with @levb, I have decided to move the postAPI interface from common to the poster package.

I am not a big fan of code repetition, and having these kind of interfaces in the "client of the interface" seemed unneeded from my point of view. But after a discussion, I understand this interface as a "import statement", in the sense of "I will be using these functions from this package".

Pros?
- Only changes in the signature of the "imported functions" will require changes on these files. If we were to have a full interface, any change on the api signature will trigger an error in this interface.
- Better knowledge of what changes in the api will affect this package (we know a change in how we implement GetPostBefore will not affect this package, since that function is not "imported").

Cons?
- Minor potential code repetition across packages, since each will need their import and their import mocks.

The main downside is the file name. I have mixed feelings about naming it import. On one side, I feel this interface to be something similar to a import statement, saying what functions we use from a different package, but at the same time `import` is a word loaded with lots of meaning, so it might not be the best word to use here.

@lieut-data thoughts on this?

lieut-data:
Agree with moving the interface -- the narrow scope of the interface required by the poster package is definitely a concern of the poster package itself.

levb:
Agree that naming can be improved (not just the filename, I cringe at using API in the type name) - 0/5 let things stabilize first, then do a round of naming cleanup?

larkox:
Understood. @levb whenever we have your review, we can merge :wink:

PR #61, opened 2020-07-08
[GH-14939] Set defaults on API Config

Szymongib:
#### Summary

Add calls to `Config.SetDefaults()` before returning configuration in `ConfigurationService`, to prevent config from containing `nil` values.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-server/issues/14939


lieut-data:
Thanks, @Szymongib! Just adding one dev review for now, as I see you've marked it as a draft. Once you're ready for a final review, we can add the second reviewer and get your improvements merged :)

PR #62, opened 2020-07-08
To be consistent with the rest of Mattermost telemetry, the user's ID…

Adovenmuehle:
… should be captured as 'UserActualID'.

#### Summary
To be consistent with the rest of Mattermost telemetry, the user's ID should be captured as 'UserActualID'.

### DIFF
--- a/experimental/telemetry/tracker.go
+++ b/experimental/telemetry/tracker.go
@@ -93,6 +93,6 @@ func (t *tracker) TrackEvent(event string, properties map[string]interface{}) {
 }
 
 func (t *tracker) TrackUserEvent(event, userID string, properties map[string]interface{}) {
-	properties["UserID"] = userID
+	properties["UserActualID"] = userID
 	t.TrackEvent(event, properties)
 }


experimental/telemetry/tracker.go
```diff
@@ -93,6 +93,6 @@ func (t *tracker) TrackEvent(event string, properties map[string]interface{}) {
 }
 
 func (t *tracker) TrackUserEvent(event, userID string, properties map[string]interface{}) {
-	properties["UserID"] = userID
+	properties["UserActualID"] = userID
```
hanzei, line 96: NPS uses `user_actual_id'. Should that also be used here? https://github.com/mattermost/mattermost-plugin-nps/blob/75b671a998901e00253fe04b840047eccb74e3a4/server/segment.go#L61

Adovenmuehle: RudderStack ends up transforming `UserActualID` into `user_actual_id` so in effect, `UserActualID` == `user_actual_id`

larkox:
@DHaussermann merging without QA Review, since it is a package on the experimental package.

experimental/telemetry/tracker.go
hanzei: Does it make sense to use `user_actual_id` here for clarity?

PR #63, opened 2020-07-13
Add DM to post api

larkox:
#### Summary
Add the DM helper function to the PostService.

#### Ticket Link
None

### DIFF
--- a/post.go
+++ b/post.go
@@ -30,18 +30,14 @@ func (p *PostService) CreatePost(post *model.Post) error {
 // DM sends a post as a direct message
 //
 // Minimum server version: 5.2
-func (p *PostService) DM(senderID, receiverID string, post *model.Post) (string, error) {
+func (p *PostService) DM(senderUserID, receiverUserID string, post *model.Post) error {
 	channel, appErr := p.api.GetDirectChannel(senderID, receiverID)
 	if appErr != nil {
 		return "", normalizeAppErr(appErr)
 	}
 	post.ChannelId = channel.Id
 	post.UserId = senderID
-	sentPost, appErr := p.api.CreatePost(post)
-	if appErr != nil {
-		return "", normalizeAppErr(appErr)
-	}
-	return sentPost.Id, nil
+	return p.CreatePost(post)
 }
 
 // GetPost gets a post.
--- a/post.go
+++ b/post.go
@@ -27,6 +27,23 @@ func (p *PostService) CreatePost(post *model.Post) error {
 	return nil
 }
 
+// DM sends a post as a direct message
+//
+// Minimum server version: 5.2
+func (p *PostService) DM(senderID, receiverID string, post *model.Post) (string, error) {
+	channel, appErr := p.api.GetDirectChannel(senderID, receiverID)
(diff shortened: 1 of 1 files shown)


larkox:
One thing that I saw and that we may want to change is also on the "CreatePost" function. We should return at least the ID of the post. Sometimes we want to create a post and keep it in check for future changes (e.g. Zoom meeting ended changes). If we were not to return the post id, we cannot reliably get back that post to do the changes we need.

post.go
```diff
@@ -27,6 +27,23 @@ func (p *PostService) CreatePost(post *model.Post) error {
...
+	if appErr != nil {
+		return "", normalizeAppErr(appErr)
+	}
+	post.ChannelId = channel.Id
+	post.UserId = senderID
+	sentPost, appErr := p.api.CreatePost(post)
+	if appErr != nil {
+		return "", normalizeAppErr(appErr)
+	}
+	return sentPost.Id, nil
```
lieut-data, line 44: This signature deviates from `CreatePost`, where the incoming `*model.Post` is considered mutable and overwritten with the created post (effectively assigning the `Id` back to the original post passed in.)

There was some discussion about this back in https://community-daily.mattermost.com/core/channels/plugin-api-rework-hackathon-2019, though I don't recall the particulars anymore. (@hanzei might remember?). I'm open to revisiting that discussion as necessary, but for now should we keep `DM` consistent with `CreatePost`?

post.go
```diff
@@ -27,6 +27,23 @@ func (p *PostService) CreatePost(post *model.Post) error {
 	return nil
 }
 
+// DM sends a post as a direct message
+//
+// Minimum server version: 5.2
+func (p *PostService) DM(senderID, receiverID string, post *model.Post) (string, error) {
```
lieut-data, line 33: nit: would `senderUserId` and `receiverUserId` be slightly more precise?

post.go
```diff
@@ -27,6 +27,23 @@ func (p *PostService) CreatePost(post *model.Post) error {
...
+	channel, appErr := p.api.GetDirectChannel(senderID, receiverID)
+	if appErr != nil {
+		return "", normalizeAppErr(appErr)
+	}
+	post.ChannelId = channel.Id
+	post.UserId = senderID
```
lieut-data, line 39: While unorthodox, some plugin bots do post /into/ DMs, e.g. if a bot is invited to a two-person conversation to help with some automation. I'm not necessarily proposing that we support this explicitly here, but just wanted to note that we there can technically be three users involved in a DM, especially from a plugin's perspective.

post.go
larkox: Sure. If the incoming *model.Post is mutated, we can follow the same pattern.

post.go
larkox: I wonder if it makes sense that being a completely different function, because you are no longer sending a DM, you are writing in a channel that happens to be a DM. If we were to put this into the DM function, most of the times we call DM would be something like:
```
DM(sender, sender, receiver, post)
```

larkox:
@DHaussermann merging without testing.

PR #65, opened 2020-08-10
Add fallback to slack attachments and add missing comments

larkox:
#### Summary
Add fallback to slack attachments messages from panel and flow.

#### Ticket Link
None

### DIFF
--- a/experimental/freetextfetcher/handler.go
+++ b/experimental/freetextfetcher/handler.go
@@ -11,13 +11,19 @@ import (
 )
 
 const (
-	ContextActionKey  = "id"
+	// ContextActionKey defines the key used in the context to store the action ID
+	ContextActionKey = "id"
+	// ContextMessageKey defines the key used in the context to store the message
 	ContextMessageKey = "message"
+	// ContextPayloadKey defines the key used in the context to store the Payload
 	ContextPayloadKey = "payload"
 
+	// ContextNewMessage defines the key used in the context to store the new message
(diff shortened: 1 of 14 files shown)


experimental/freetextfetcher/handler.go
```diff
@@ -11,13 +11,19 @@ import (
 )
 
 const (
-	ContextActionKey  = "id"
+	// ContextActionKey defines the key used in the context to store the action ID
+	ContextActionKey = "id"
+	// ContextMessageKey defines the key used in the context to store the message
 	ContextMessageKey = "message"
+	// ContextPayloadKey defines the key used in the context to store the Payload
```
jfrerich, line 18:
```suggestion
	// ContextPayloadKey defines the key used in the context to store the payload
```

jfrerich: suggest lowercasing to match other comments

larkox:
I am not 100% familiar with the field neither, but according to the [documentation](https://docs.mattermost.com/developer/message-attachments.html):
> fallback: A required plain-text summary of the attachment. This is used in notifications, and in clients that don’t support formatted text (e.g. IRC).

The main use from my side was to try to fix https://github.com/mattermost/mattermost-plugin-mscalendar/issues/169

If this works as expected, if by any chance a generated message triggers a mail notification, the text will have something to show. The same way, people using other clients for mattermost that do not support attachments will show the message properly.

Issue #14, opened 2019-11-27
Add i18n service


hanzei:
Might be inspired by https://github.com/matterpoll/matterpoll/blob/master/server/plugin/i18n.go

hanzei:
See https://github.com/matterpoll/matterpoll/issues/260 for discussion on how the api should look.

PR #66, opened 2020-08-13
Add i18n package

hanzei:
#### Summary
This package provides the needed tools to allow plugins to quickly add support for i18n.

I chose too alias most of they `types` from github.com/nicksnyder/go-i18n/v2 to prevent consumers of this library to import them there self. `Bundle` is composed of `go-i18n.Bundle`, which allows for flexible use. To simplify the API, `Bundle.Bundle` could be made private.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-api/issues/14

### DIFF
--- a/i18n/i18n_test.go
+++ b/i18n/i18n_test.go
@@ -1,0 +1,296 @@
+package i18n_test
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mattermost-server/v5/plugin"
+	"github.com/mattermost/mattermost-server/v5/plugin/plugintest"
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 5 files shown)


i18n/i18n_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+func ExampleInitBundle() {
+	type Plugin struct {
+		plugin.MattermostPlugin
+
+		b *i18n.Bundle
+	}
+
+	p := Plugin{}
+	b, err := i18n.InitBundle(p.API, filepath.Join("assets", "i18n"))
```
larkox, line 28: Just wondering whether we should promote using the plugin api client defined in this repo. Maybe it makes things quite complicated in terms of interfacing the API needed, but it may make sense.

hanzei: 1/5 I feel like the api client isn't ready to be used yet. There a still a couple of issues to resolve first. But maybe this topic only needs a final push.

PR #69, opened 2020-09-07
Add telemetry client events handler

larkox:
#### Summary
Add telemetry client events handler

#### Ticket Link
None

### DIFF
--- a/experimental/telemetry/tracker.go
+++ b/experimental/telemetry/tracker.go
@@ -126,7 +126,10 @@ func (t *tracker) HandleClientEvent(r *http.Request) error {
 	}
 
 	if telemetryRequest.Event != "" {
-		_ = t.TrackUserEvent(telemetryRequest.Event, userID, telemetryRequest.Properties)
+		err = t.TrackUserEvent(telemetryRequest.Event, userID, telemetryRequest.Properties)
+		if err != nil {
+			return errors.Wrap(err, "unable to track event")
+		}
 	}
 
 	return nil
--- a/experimental/telemetry/tracker.go
+++ b/experimental/telemetry/tracker.go
@@ -1,6 +1,9 @@
 package telemetry
 
 import (
+	"encoding/json"
+	"net/http"
+
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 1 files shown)


experimental/telemetry/tracker.go
```diff
@@ -105,3 +108,28 @@ func (t *tracker) TrackUserEvent(event, userID string, properties map[string]int
...
+func (t *tracker) HandleClientEvent(w http.ResponseWriter, r *http.Request) {
+	userID := r.Header.Get("Mattermost-User-ID")
+	if userID == "" {
+		http.Error(w, "Not authorized", http.StatusUnauthorized)
+		return
+	}
+
+	var telemetryRequest *telemetryAPIRequest
+	decoder := json.NewDecoder(r.Body)
+	err := decoder.Decode(&telemetryRequest)
```
hanzei, line 126: Nit:
```suggestion
	err := json.NewDecoder(r.Body).Decode(&telemetryRequest)
```

experimental/telemetry/tracker.go
```diff
@@ -105,3 +108,28 @@ func (t *tracker) TrackUserEvent(event, userID string, properties map[string]int
...
+	if err != nil {
+		http.Error(w, "Unable to decode JSON", http.StatusBadRequest)
```
hanzei, line 128: It's a bummer that there is no entity to log the error

experimental/telemetry/tracker.go
```diff
@@ -105,3 +108,28 @@ func (t *tracker) TrackUserEvent(event, userID string, properties map[string]int
...
+		return
+	}
+
+	if telemetryRequest.Event != "" {
+		_ = t.TrackUserEvent(telemetryRequest.Event, userID, telemetryRequest.Properties)
```
hanzei, line 133: Same here

hanzei (approved):
LGTM :+1: 

I would love to log the errors, but that is out of scope of this PR

experimental/telemetry/tracker.go
larkox: What we can do is to return the error, and leave the http handling to the caller. What do you think?

hanzei: Leaving it to the caller would push the error to the webapp in most of the cases. For debugging it's more useful to log the error in the server logs.

larkox: Right now the tracker does not have "logging functions", and I would prefer to leave it like that. I made some changes to return the error.

experimental/telemetry/tracker.go
```diff
@@ -105,3 +107,30 @@ func (t *tracker) TrackUserEvent(event, userID string, properties map[string]int
 	properties["UserActualID"] = userID
 	return t.TrackEvent(event, properties)
 }
+
+type telemetryAPIRequest struct {
+	Event      string
+	Properties map[string]interface{}
+}
+
+func (t *tracker) HandleClientEvent(r *http.Request) error {
```
hanzei, line 116: I'm not sure if that signature is better. With `(w http.ResponseWriter, r *http.Request)` is was easily plugable and with that change a consumer as to write there own error handling method. I would like to keep the methods a just plug an play.

PR #72, opened 2020-09-08
allow development mode when checking license

lieut-data:
#### Summary
Port the changes from incident response to skip the license check when `EnableDeveloper` and `EnableTesting` is true. These bits are not suitable for use in production, and allow for development testing under the source available license without having to run the enterprise server binary.

Once merged, I'll leverage same in incident response both to simplify the code there and allow our use of the SQL store under the same circumstances.

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-28104

### DIFF
--- a/license.go
+++ b/license.go
@@ -1,0 +1,36 @@
+package pluginapi
+
+import (
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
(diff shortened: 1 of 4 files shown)


ali-farooq0 (approved):
👌

cpoile (approved):

PR #74, opened 2020-09-17
Add WithError method to Logger

hanzei:
#### Summary
That allows for easier logging of error in an idiomatic way.

#### Ticket Link
None


larkox (approved):
LGTM. Really nice solution :D

Issue #22, opened 2019-11-27
Fix 404 error situations


hanzei:
The expected behaviour is:
- For `Get` methods: Return `(nil, 404)` if resource doesn't exist.
- For `List` methods: Return `(nil, nil)` if resource doesn't exist.

PR #75, opened 2020-09-23
return ErrNotFound when normalizing errors

lieut-data:
#### Summary
Simplify handling of errors by converging on a common `StatusNotFound` error instead of having to discover if the returned `error` is an `*model.AppError` and examine the `StatusCode`.

Example usage:
```go
	user, err := r.pluginAPI.User.GetByUsername(username)
	if errors.Is(err, pluginapi.ErrNotFound) {
		pluginAPI.Log.Info(fmt.Sprintf("Unable to find user @%s", username)))
		return
	} else if err != nil {
		pluginAPI.Log.Warn(fmt.Sprintf("Unable to find user @%s", username)), "err", err)
		return
	}
```

### DIFF
--- a/error.go
+++ b/error.go
@@ -1,10 +1,17 @@
 package pluginapi
 
 import (
+	"net/http"
+
 	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/pkg/errors"
 )
 
-// normalizeAppError returns a truly nil error if appErr is nil.
+// ErrNotFound is returned by the plugin API when an object is not found.
+var ErrNotFound = errors.New("not found")
+
+// normalizeAppError returns a truly nil error if appErr is nil as well as normalizing a class
+// of non-nil AppErrors to simplify use within plugins.
 //
 // This doesn't happen automatically when a *model.AppError is cast to an error, since the
 // resulting error interface has a concrete type with a nil value. This leads to the seemingly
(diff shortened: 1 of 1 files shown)


error.go
```diff
@@ -23,5 +30,9 @@ func normalizeAppErr(appErr *model.AppError) error {
 		return nil
 	}
 
+	if appErr.StatusCode == http.StatusNotFound {
+		return ErrNotFound
+	}
```
hanzei, line 35: In the discussion about error handling we decided to not return an error for `List` methods that return an empty list. https://github.com/mattermost/mattermost-plugin-api/issues/22 encapsulate that decisions.

Are we sticking with the decision? If so, can you confirm that no `List` method returns an error?

hanzei (changes requested):
Great work. That will make development easier :+1: One concern about List methods

error.go
lieut-data: Yes, I'm inclined to stick with that decision -- an empty list is not an error. Do we need to comb through anything in advance of formally addressing https://github.com/mattermost/mattermost-plugin-api/issues/22? If any `List` methods return errors today instead of empty results, this doesn't make that /worse/, strictly speaking.

agarciamontoro (approved):
Neat! :tada:

error.go
hanzei: Good point, this PR doesn't make it worse. With that in mind I'm fine with merging.

Yes, we have to comb trough every List method for #22.

hanzei (approved):

PR #76, opened 2020-09-28
Fixed JSON unmarshaling in experimental/oAuther

levb:
#### Summary
Fixed JSON unmarshaling in experimental/oAuther

#### Ticket Link
none

### DIFF
--- a/experimental/oauther/oauth2.go
+++ b/experimental/oauther/oauth2.go
@@ -142,7 +142,7 @@ func (o *oAuther) GetConnectURL() string {
 
 func (o *oAuther) GetToken(userID string) (*oauth2.Token, error) {
 	var token *oauth2.Token
-	err := o.store.Get(o.getTokenKey(userID), token)
+	err := o.store.Get(o.getTokenKey(userID), &token)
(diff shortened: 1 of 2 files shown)


experimental/oauther/oauth2.go
```diff
@@ -142,7 +142,7 @@ func (o *oAuther) GetConnectURL() string {
 
 func (o *oAuther) GetToken(userID string) (*oauth2.Token, error) {
 	var token *oauth2.Token
```
larkox, line 144: Should we here then use the object directly instead of a pointer?

larkox (approved):
LGTM, just one question

experimental/oauther/oauth2.go
hanzei: Good suggestion :+1:

hanzei (approved):

experimental/oauther/oauth2.go
levb: It kinda doesn't matter, `json` will do the right thing, but it needs a *pointer* to the output variable

PR #77, opened 2020-10-21
Revert "[GH-14939] Set defaults on API Config (#61)"

cpoile:
#### Summary
- Since gob sometimes turns `false`s into `nil`s (https://mattermost.atlassian.net/browse/MM-14595), we thought we should apply the defaults in #61 . 
- That turned out to cause an unexpected behavior: some config values set to false (e.g., `LogSettings.EnableDiagnostics`) will never be false, they will always have the default of true. This is unexpected and caused Christopher and I a bit of pain today to track down why we couldn't turn off telemetry.
- Plugins should be checking whether a field is nil before dereferencing it. If it's nil, treat it as a false.


crspeller (approved):

agarciamontoro (approved):
Oh no, that's inconvenient! Approving this now, but what if instead of using `SetDefaults` we use a custom `CleanNils` that applies the default value when needed and false when there is a boolean value?

> Plugins should be checking whether a field is nil before dereferencing it. If it's nil, treat it as a false.

:sob:

cpoile:
@agarciamontoro Unfortunately, that would result in the same problem we're having now -- nils should be false, not the default. So I suppose we could make CleanNils apply false values instead of the default. Let's leave that for a HW then.

agarciamontoro:
> So I suppose we could make CleanNils apply false values instead of the default.

Yup, that's what I meant, sorry for the confusion :)

> Let's leave that for a HW then.

Sounds good!

cpoile:
@agarciamontoro ahh, yes, I see that now -- I misread your comment. Sorry!  :)

PR #78, opened 2020-10-25
fix broken README link

jfrerich:
#### Summary
Navigating to the given link produces a 404. This PR keeps the text shown to the user the same but fixes the destination link.

#### Ticket Link
n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 A hackathon project to explore reworking the Mattermost Plugin API.
 
-The plugin API as exposed in [github.com/mattermost/mattermost-server/plugin](http://github.com/mattermost/mattermost-server/plugin) began with the hope of adopting a consistent interface and style. But our vision for how to structure the API changed over time, along with our ability to remain consistent. 
+The plugin API as exposed in [github.com/mattermost/mattermost-server/plugin](http://github.com/mattermost/mattermost-server/tree/master/plugin) began with the hope of adopting a consistent interface and style. But our vision for how to structure the API changed over time, along with our ability to remain consistent. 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,7 +6,7 @@
 
 A hackathon project to explore reworking the Mattermost Plugin API.
 
-The plugin API as exposed in [github.com/mattermost/mattermost-server/plugin](http://github.com/mattermost/mattermost-server/plugin) began with the hope of adopting a consistent interface and style. But our vision for how to structure the API changed over time, along with our ability to remain consistent. 
+The plugin API as exposed in [github.com/mattermost/mattermost-server/plugin](http://github.com/mattermost/mattermost-server/tree/master/plugin) began with the hope of adopting a consistent interface and style. But our vision for how to structure the API changed over time, along with our ability to remain consistent.
```
hanzei, line 9: Should we even link to https://github.com/mattermost/mattermost-server/blob/master/plugin/api.go instead?

hanzei (changes requested):
Thanks for fixing this :+1:

README.md
jfrerich: I think that makes the most sense and will adopt your suggestion @hanzei.  We could also point to the https://developers.mattermost.com/extend/plugins/server/reference/#API, or both?

hanzei: Let's point to the original source at https://github.com/mattermost/mattermost-server/blob/master/plugin/api.go?rgh-link-date=2020-10-26T12%3A12%3A01Z

lieut-data (approved):

hanzei (approved):

PR #83, opened 2021-02-19
Relax store license restrictions

lieut-data:
#### Summary
When we designed the store access in the pluginapi, we made the conscious decision to enforce an E20 license (outside of developer environments).

Unfortunately, this limitation has caused us trouble in Incident Collaboration, where we want to be able to start up the plugin even without a valid E20 license, and selectively enable certain functionality for non-E20 licensed customers. Since the pluginapi enforces E20, we can't get our base functionality off the ground.

After further review, we have decided to simply [lift this restriction altogether](https://community-daily.mattermost.com/core/pl/3akxk61ebjfqmbbh588pn5mrjr). It was always possible to work around anyway.

#### Ticket Link
None.

### DIFF
--- a/store.go
+++ b/store.go
@@ -14,9 +14,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-// ErrNoLicense is returned only when there is no enterprise license set.
-var ErrNoLicense = errors.New("this feature requires a valid enterprise license")
-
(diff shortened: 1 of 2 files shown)


mickmister (approved):
LGTM!

catalintomai (approved):

PR #88, opened 2021-04-30
Add version-aware RequestTrialLicense to System

agarciamontoro:
#### Summary
Adds a RequestTrialLicense that gracefully degrades in servers with a version lower to 5.36.0

This needs to wait for https://github.com/mattermost/mattermost-server/pull/17551 to be merged.

#### Ticket Link
Needed by https://mattermost.atlassian.net/browse/MM-34665

### DIFF
--- a/system.go
+++ b/system.go
@@ -74,7 +74,7 @@ func (s *SystemService) GetDiagnosticID() string {
 // RequestTrialLicense requests a trial license and installs it in the server.
 // If the server version is lower than 5.36.0, an error is returned.
 //
-// Minimum server version: 5.4
+// Minimum server version: 5.36
 func (s *SystemService) RequestTrialLicense(requesterID string, users int, termsAccepted, receiveEmailsAccepted bool) error {
 	currentVersion := semver.MustParse(s.api.GetServerVersion())
 	requiredVersion := semver.MustParse("5.36.0")
--- a/system.go
+++ b/system.go
@@ -3,6 +3,7 @@ package pluginapi
 import (
 	"time"
 
+	"github.com/blang/semver/v4"
 	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/mattermost/mattermost-server/v5/plugin"
 	"github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


agarciamontoro:
Not sure why the `job_once_test` failed here. Any ideas @cpoile?

system.go
```diff
@@ -69,3 +70,19 @@ func (s *SystemService) GetDiagnosticID() string {
...
+// If the server version is lower than 5.36.0, an error is returned.
+//
+// Minimum server version: 5.4
+func (s *SystemService) RequestTrialLicense(requesterID string, users int, termsAccepted, receiveEmailsAccepted bool) error {
+	currentVersion := semver.MustParse(s.api.GetServerVersion())
+	requiredVersion := semver.MustParse("5.36.0")
+
+	if currentVersion.LT(requiredVersion) {
+		return errors.Errorf("current server version is lower than 5.36")
+	}
```
agarciamontoro, line 84: Is this how this should be done? Is there any better way? I'm all ears.

system.go
```diff
@@ -69,3 +70,19 @@ func (s *SystemService) GetDiagnosticID() string {
 	// TODO: Consider deprecating/rewriting in favor of just using GetUnsanitizedConfig().
 	return s.api.GetDiagnosticId()
 }
+
+// RequestTrialLicense requests a trial license and installs it in the server.
+// If the server version is lower than 5.36.0, an error is returned.
+//
+// Minimum server version: 5.4
```
agarciamontoro, line 77: I set this to the minimum server version of GetServerVersion, which is the only API call we do without actually checking the server version before.

system.go
lieut-data: What happens if we try to invoke `RequestTrialLicense` without a server implementation? Is that error code predictable, and usable without having to explicitly check versions?

lieut-data (approved):

system.go
agarciamontoro: I have not actually checked, but I think I like having the explicit version check. It signals the intention better than handling an error, which may need the addition of a comment to explain what's going on.

system.go
lieut-data: Is that useful, though? Seems like having a floor on which we can tell if an error occurred isn't what the consumer of this API might want, but when the API is likely to be successful.

cpoile: To be clear: You're suggesting we require 5.36.0 and don't worry about the manual version test?

cpoile (approved):
Approving, contingent on Jesse's concerns. Nice work @agarciamontoro!

system.go
lieut-data: No, I think it's fine to have the manual test and return an explicit error to the client that this API isn't supported. In fact, I'd love for all the plugin API wrappers to effectively allow this, making it easy to test distinguish downstream.

I'm just objecting to the documentation showing this as v5.4:
* From v5.2-v5.4, this API will fail cryptically
* From v5.4-v5.35, this API will fail with a hard-coded error
* From v5.36, this API will return the desired result.

I /feel/ like v5.36 is the version we're intending to communicate here, not that we successfully avoid a cryptic error.

agarciamontoro:
To be tested as part of https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/546

PR #91, opened 2021-06-15
MM-36167: Use SKU short name with feature check as fallback

agarciamontoro:
#### Summary
This PR updates the  license checks to use the proposed new way of detecting them. The details are [in this document](https://docs.google.com/document/d/1PhlIiGXbAl-sTRFXLRT5R71eSzUTA0HXAVNXPu1aAVU/edit), but the gist of it is that we are now using the SKU shortname instead of specific features. As the old names will be in use for at least two more years, we need to check for both the old ones (E10, E20) and the new ones (professional, enterprise). The possible names are the keys in this map: https://github.com/mattermost/customer-web-server/blob/master/licenses/licenses.go#L98

As a fallback for legacy licenses with potentially unknown SKU shortnames, we default to the old way of looking at a specific feature that we know is in the license we are checking. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36167

### DIFF
--- a/license.go
+++ b/license.go
@@ -4,6 +4,13 @@ import (
 	"github.com/mattermost/mattermost-server/v5/model"
 )
 
+const (
+	e10          = "E10"
+	e20          = "E20"
+	professional = "professional"
+	enterprise   = "enterprise"
+)
+
 // IsEnterpriseLicensedOrDevelopment returns true when the server is licensed with any Mattermost
 // Enterprise License, or has `EnableDeveloper` and `EnableTesting` configuration settings
 // enabled signaling a non-production, developer mode.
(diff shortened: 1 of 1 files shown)


license.go
```diff
@@ -15,31 +22,64 @@ func IsEnterpriseLicensedOrDevelopment(config *model.Config, license *model.Lice
...
+		// and use the presence of a known E10/Professional feature as a check to determine licensing.
+		if license != nil &&
+			license.Features != nil &&
+			license.Features.LDAP != nil &&
+			*license.Features.LDAP {
+			return true
+		}
+	}
+
 	return IsConfiguredForDevelopment(config)
```
lieut-data, line 61: If we moved this check to the top, I wonder if this would help with the flow? (0/5)

lieut-data (approved):
Looks good, but can we add unit tests to exercise these cases?

license.go
```diff
@@ -15,31 +22,64 @@ func IsEnterpriseLicensedOrDevelopment(config *model.Config, license *model.Lice
...
+		return true
+	default:
+		return false
+	}
+}
+
+// IsE10LicensedOrDevelopment returns true when the server is at least licensed with a legacy Mattermost
+// Enterprise E10 License or a Mattermost Professional License, or has `EnableDeveloper` and
+// `EnableTesting` configuration settings enabled, signaling a non-production, developer mode.
 func IsE10LicensedOrDevelopment(config *model.Config, license *model.License) bool {
```
iomodo, line 43: You just could check `license.SkuShortName == e10 || license.SkuShortName == professional || IsE20LicensedOrDevelopment(...) ` or am I missing something?

iomodo (approved):
Looks good!
Just one non-blocking comment

PR #93, opened 2021-06-18
Add DB RPC API support to client

agnivade:
We add support for the StoreService
to use the newly added plugin.Driver interface to make
SQL queries.

This allows all plugins using the plugin-api
to route all SQL queries via Mattermost server
using the central connection pool, and avoid
having to open new DB connections.

This is implemented on the server via a database
driver which acts as an RPC wrapper, giving clients
a zero-cost abstraction on which to run their SQL queries.

All clients get a *sql.DB object as normal, but under
the hood they are routed via RPC.

https://focalboard-community.octo.mattermost.com/workspace/zyoahc9uapdn3xdptac6jb69ic?id=285b80a3-257d-41f6-8cf4-ed80ca9d92e5&v=495cdb4d-c13a-4992-8eb9-80cfee2819a4&c=c7386db7-65fd-469b-8bcf-8dc8f8e61e4f


client.go
```diff
@@ -50,9 +50,41 @@ func NewClient(api plugin.API) *Client {
...
+			api:     api,
+			version: 0,
+		},
+		System: SystemService{api: api},
+		Team:   TeamService{api: api},
+		User:   UserService{api: api},
+	}
+}
+
+func NewClient2(api plugin.API, driver plugin.Driver) *Client {
```
agnivade, line 63: @lieut-data - Looking for some guidance on this. In an ideal world, I'd like to break API altogether and avoid having two `Client` constructors and versioning in the `StoreService`. For now, I just added a `NewClient2` to start the discussion.

Let me know how to proceed.

jespino (approved):
LGTM

client.go
lieut-data: Actually, feel free to break this API all you want: we're currently <1.0 on purpose, with the assumption that consumers will upgrade intentionally and deal with the fallout then.

lieut-data (commented):
This seems pretty slick to me -- but adding @crspeller for review.

client.go
agnivade: Wonderful, thanks!

agnivade:
ping @crspeller :)

store_test.go
```diff
@@ -45,16 +43,7 @@ func TestStore(t *testing.T) {
 		require.NoError(t, store.Close())
 	})
 
-	t.Run("master db", func(t *testing.T) {
-		db, err := sql.Open("ramsql", "TestStore-master-db")
-		require.NoError(t, err)
-		defer db.Close()
-
-		_, err = db.Exec("CREATE TABLE test (id INT);")
```
agnivade, line 53: I have removed the actual SQL queries because with the RPC API, we would be actually testing the mocks. There's no benefit of testing just for the sake of it. I have only kept the master/replica verification logic.

crspeller (approved):
Great! This will make cloud happy.

PR #95, opened 2021-06-25
[MM-36664] Added plugin API methods for user access tokens and OAuth apps

hanzei:
#### Summary
Downstream changes from https://github.com/mattermost/mattermost-server/pull/17832

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-36664


larkox (approved):

oauth.go
```diff
@@ -0,0 +1,41 @@
...
+// Create creates a new OAuth App,
+//
+// Minimum server version: 5.38
+func (o *OAuthService) Create(app *model.OAuthApp) error {
+	createdApp, appErr := o.api.CreateOAuthApp(app)
+	if appErr != nil {
+		return normalizeAppErr(appErr)
+	}
+
+	*app = *createdApp
```
mickmister, line 22: It seems odd to me that we are mutating the passed in `app`, instead of returning `createdApp`. Or this may be something common in this library?

hanzei: That is in deed common in this library. It allows for shorter code like:
```go
post := &model.Post{Message: "abc", ChannelID: "someID"}
err := client.Post.Create(post)
if err != nil {
	panic(err)
}
log.Printf("Post as ID %s", post.Id)
```

Also, the pattern is consistently used in the code base and hence in this PR.

user.go
```diff
@@ -217,3 +217,24 @@ func (u *UserService) GetLDAPAttributes(userID string, attributes []string) (map
 
 	return ldapUserAttributes, normalizeAppErr(appErr)
 }
+
+// CreateAccessToken creates a new access token
```
mickmister, line 221:
```suggestion
// CreateAccessToken creates a new access token.
```

oauth.go
```diff
@@ -0,0 +1,41 @@
...
+// Get gets an existing OAuth App.
+//
+// Minimum server version: 5.38
+func (o *OAuthService) Get(appID string) (*model.OAuthApp, error) {
+	app, appErr := o.api.GetOAuthApp(appID)
+
+	return app, normalizeAppErr(appErr)
+}
+
+// Get gets an existing OAuth App by id.
```
mickmister, line 36:
```suggestion
// Delete deletes an existing OAuth App by id.
```

oauth.go
```diff
@@ -0,0 +1,41 @@
...
+	if appErr != nil {
+		return normalizeAppErr(appErr)
+	}
+
+	*app = *createdApp
+
+	return nil
+}
+
+// Get gets an existing OAuth App.
```
mickmister, line 27: 0/5 to match the comment for the `Delete` method like this

```suggestion
// Get gets an existing OAuth App by id.
```

oauth.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mattermost-server/v5/plugin"
+)
+
+// UserService exposes methods to manipulate OAuth Apps.
+type OAuthService struct {
+	api plugin.API
+}
+
+// Create creates a new OAuth App,
```
mickmister, line 13:
```suggestion
// Create creates a new OAuth App.
```

mickmister: Do we ever need to patch an existing OAuth App?

mickmister (approved):
LGTM :+1: I have some non-blocking requests on doc comments

oauth.go
hanzei: Maybe. Let me just add a method for that to be sure.

hanzei: Added an `Update` method.

PR #96, opened 2021-06-28
MM-32421: Extend channel sidebar category APIs for PluginAPI

madhavhugar:
#### Summary

Add CreateSidebarCategory, GetSidebarCategories and UpdateSidebarCategories to PluginAPI Channel Service

#### Ticket Link

JIRA: https://mattermost.atlassian.net/browse/MM-32421

#### Release Note

```release-note
Added CreateSidebarCategory, GetSidebarCategories and UpdateSidebarCategories to PluginAPI Channel Service
```

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -217,6 +217,36 @@ func (c *ChannelService) UpdateChannelMemberNotifications(channelID, userID stri
 	return channelMember, normalizeAppErr(appErr)
 }
 
+// TODO: should we create a SidebarService?
+
+// CreateSidebarCategory creates a new sidebar category for a set of channels.
+//
+// Minimum server version: 5.37
+func (c *ChannelService) CreateSidebarCategory(userID, teamID string, newCategory *model.SidebarCategoryWithChannels) (*model.SidebarCategoryWithChannels, error) {
+	category, appErr := c.api.CreateChannelSidebarCategory(userID, teamID, newCategory)
+
+	return category, normalizeAppErr(appErr)
+}
+
+// GetSidebarCategories returns sidebar categories.
+//
+// Minimum server version: 5.37
+func (c *ChannelService) GetSidebarCategories(userID, teamID string) (*model.OrderedSidebarCategories, error) {
+	categories, appErr := c.api.GetChannelSidebarCategories(userID, teamID)
+
+	return categories, normalizeAppErr(appErr)
+}
+
+// UpdateSidebarCategories updates the channel sidebar categories.
+//
+// Minimum server version: 5.37
+func (c *ChannelService) UpdateSidebarCategories(userID, teamID string, categories []*model.SidebarCategoryWithChannels) ([]*model.SidebarCategoryWithChannels, error) {
+	updatedCategory, appErr := c.api.UpdateChannelSidebarCategories(userID, teamID, categories)
+
+	return updatedCategory, normalizeAppErr(appErr)
+}
+
+
 func (c *ChannelService) waitForChannelCreation(channelID string) error {
 	if len(c.api.GetConfig().SqlSettings.DataSourceReplicas) == 0 {
 		return nil
--- a/channel.go
+++ b/channel.go
@@ -221,10 +221,14 @@ func (c *ChannelService) UpdateChannelMemberNotifications(channelID, userID stri
 //
 // Minimum server version: 5.37
 func (c *ChannelService) CreateSidebarCategory(
-	userID, teamID string, newCategory *model.SidebarCategoryWithChannels) (*model.SidebarCategoryWithChannels, error) {
+	userID, teamID string, newCategory *model.SidebarCategoryWithChannels) error {
 	category, appErr := c.api.CreateChannelSidebarCategory(userID, teamID, newCategory)
+	if appErr != nil {
+		return normalizeAppErr(appErr)
+	}
+	*newCategory = *category
 
-	return category, normalizeAppErr(appErr)
+	return nil
 }
 
 // GetSidebarCategories returns sidebar categories.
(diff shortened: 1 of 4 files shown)


madhavhugar:
@agarciamontoro created this PR to support new Plugin APIs for [categorizing incident channel](https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/598). Please review.

agarciamontoro:
Thank you, @madhavhugar, we'll review this as soon as we can!

lieut-data (approved):
Nice! Thanks again, @madhavhugar :)

channel.go
```diff
@@ -217,6 +217,35 @@ func (c *ChannelService) UpdateChannelMemberNotifications(channelID, userID stri
 	return channelMember, normalizeAppErr(appErr)
 }
 
+// CreateSidebarCategory creates a new sidebar category for a set of channels.
+//
+// Minimum server version: 5.37
+func (c *ChannelService) CreateSidebarCategory(
+	userID, teamID string, newCategory *model.SidebarCategoryWithChannels) (*model.SidebarCategoryWithChannels, error) {
```
hanzei, line 224: The general design guideline for the API has been to not return an point to the object, which was created and instead overwrite the passed pointer as in https://github.com/mattermost/mattermost-plugin-api/blob/e6371ea3fdddd63a2d436be72625424accbd1241/user.go#L93-L99. Is the a reason this PR deviates from that pattern?

agarciamontoro: I did not know about this convention, but I'm all in for consistency. @madhavhugar, do you mind following the pattern suggested by @hanzei?

channel.go
```diff
@@ -217,6 +217,35 @@ func (c *ChannelService) UpdateChannelMemberNotifications(channelID, userID stri
...
+
+	return categories, normalizeAppErr(appErr)
+}
+
+// UpdateSidebarCategories updates the channel sidebar categories.
+//
+// Minimum server version: 5.37
+func (c *ChannelService) UpdateSidebarCategories(
+	userID, teamID string, categories []*model.SidebarCategoryWithChannels) ([]*model.SidebarCategoryWithChannels, error) {
+	updatedCategory, appErr := c.api.UpdateChannelSidebarCategories(userID, teamID, categories)
```
agarciamontoro, line 244: We should follow the same pattern here.

agarciamontoro (changes requested):
Again, fantastic! Thank you, @madhavhugar, I only echoed what @hanzei suggested about the pattern we're following in this repository.

agarciamontoro (commented):
By the way, I think the linter is complaining because you need to run `go mod tidy`.

channel.go
madhavhugar: @hanzei @agarciamontoro the pattern makes sense to reduce memory footprint, I've made the changes

hanzei (approved):
LGTM

channel.go
```diff
@@ -217,6 +217,43 @@ func (c *ChannelService) UpdateChannelMemberNotifications(channelID, userID stri
...
+// Minimum server version: 5.37
+func (c *ChannelService) UpdateSidebarCategories(
+	userID, teamID string, categories *[]*model.SidebarCategoryWithChannels) error {
+	updatedCategories, appErr := c.api.UpdateChannelSidebarCategories(userID, teamID, *categories)
+	if appErr != nil {
+		return normalizeAppErr(appErr)
+	}
+	*categories = updatedCategories
+
+	return nil
```
agarciamontoro, line 254: We can probably remove the pointer here, as the slice is already a pointer, and simply copy the contents. Thoughts, @hanzei?

```suggestion
func (c *ChannelService) UpdateSidebarCategories(
	userID, teamID string, categories []*model.SidebarCategoryWithChannels) error {
	updatedCategories, appErr := c.api.UpdateChannelSidebarCategories(userID, teamID, categories)
	if appErr != nil {
		return normalizeAppErr(appErr)
	}
	copy(categories, updatedCategories)

	return nil
```

agarciamontoro (changes requested):
Just something to make the code a bit more readable, but looks great otherwise!

channel.go
madhavhugar: looks good, but can we be sure that the length of `categories` and `updatedCategories` is always the same?

agarciamontoro: Right now, it is. You can check the code in the channel store here: https://github.com/mattermost/mattermost-server/blob/master/store/sqlstore/channel_store_categories.go#L747-L748 I think it's safe enough to assume it will be like this in the foreseeable future; otherwise, it would be a breaking change.

channel.go
```diff
@@ -217,6 +217,43 @@ func (c *ChannelService) UpdateChannelMemberNotifications(channelID, userID stri
...
+func (c *ChannelService) UpdateSidebarCategories(
+	userID, teamID string, categories []*model.SidebarCategoryWithChannels) error {
+	updatedCategories, appErr := c.api.UpdateChannelSidebarCategories(userID, teamID, categories)
+	if appErr != nil {
+		return normalizeAppErr(appErr)
+	}
+	copy(categories, updatedCategories)
+
+	return nil
+}
```
agarciamontoro, line 255: @hanzei, just to make sure: does this fit the expected pattern of updating the function parameter?

agarciamontoro (approved):
Looks great, @madhavhugar, thank you very much :)

channel.go
hanzei: Jep, that works for me.

hanzei (approved):

agarciamontoro:
QA review will happen as part of https://github.com/mattermost/mattermost-plugin-incident-collaboration/pull/598, merging now! Thanks again for your contribution, @madhavhugar :)

PR #102, opened 2021-08-24
MM-37539 - Fix a job_once bug

cpoile:
#### Summary
- In a cluster, suppose a plugin removed an old `JobOnce` and then made a new one with the same key, with a different `runAt` value. It was possible for another node to fire the original `JobOnce`, and seeing that the key still retrieved `metadata` from the KVStore, it thought the original (cancelled) `JobOnce` was still valid and would call the plugin. 
- To fix this, check the retrieved metadata to see if it is the same as the data in the `JobOnce` struct. Since the key is the same, check to see if the `runAt` value is still the same. If it's different, that means the original `JobOnce` has been rescheduled.

#### NOTE
- I will cherry pick this to a temp v5 branch after it's merged to master. We will use that temp v5 commit for `Playbooks` v1.18.0

#### Ticket Link
- https://mattermost.atlassian.net/browse/MM-37539

### DIFF
--- a/cluster/job_once.go
+++ b/cluster/job_once.go
@@ -124,8 +124,8 @@ func (j *JobOnce) run() {
 				return
 			}
 
-			// If key doesn't exist, the job has been completed already
-			if metadata == nil {
+			// If key doesn't exist, or if the runAt has changed, the original job has been completed already
+			if metadata == nil || !j.runAt.Equal(metadata.RunAt) {
 				j.cancelWhileHoldingMutex()
 				return
 			}


lieut-data (changes requested):
Makes sense! But can we update the `job_once_test.go` to exercise this case?

agarciamontoro (approved):
Great work on investigating this, @cpoile :) Thank you!

cpoile:
@lieut-data I've done my best to simulate the bug this will prevent. And indeed, the new test fails when you revert the PR's fix, so I think the test is working.

lieut-data (approved):
Clever! Thanks, @cpoile :)

PR #107, opened 2021-12-10
[MM-39951] Guard ensureBot with mutex

hanzei:
#### Summary
As discussed in https://community-daily.mattermost.com/core/pl/uwbz8nro8tywfnxjj9bw8ernma the execution of `ensureBot` is now guarded by a mutex to ensure two instances of the plugin don't race to create the bot.
- 9125e7dc7d3dff313d125ebcd47f18b2284927fa adds the lock to `ensurebot` and move the tests in `bot_test.go` from `pluginapi_test` to `pluginapi`. Given that the root package now depends on `cluster`, I've moved `ErrNotFound` to a new `errors` package, as both the root package as `cluster` use it.
- ef49dec122ea8587bd9ac006c12a92806fef3f73 pins the mockgen version so `make mock` can be run with newer go versions again.
- 35d9078319a9a5011bdea58985d2f0b4bdd8d183 removes an unneeded fail save `defer` statement from ensureBot` as the mutex now ensure that only plugin executes `ensureBot` as a time.
- 8e21251fde0f8f145fdc4af5c53364795d298708 cleans up the `ensureBot` code a bit by using idiomatic length checks for strings and unify error variables naming.
- cf4d282da26871d96e0f1df03defe5de4ece0e94 move the code from `EnsureBot` code into `bot.go` as the tests also live in `bot_test.go`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-39951


lieut-data:
Looping in @iomodo for a review, as I won't be able to give this one the attention it needs right now. Thanks @hanzei!

iomodo (approved):
Looks great Ben! Thanks!

cpoile (approved):
Looks great, thanks Ben!

hanzei:
@DHaussermann This should be tested alongside https://github.com/mattermost/mattermost-plugin-api/pull/107

PR #108, opened 2021-12-10
Migrate away from using io/ioutil

hanzei:
#### Summary
`io/ioutil` is deprecated, so let's use the appropriate packages instead

#### Ticket Link
None

### DIFF
--- a/bot_test.go
+++ b/bot_test.go
@@ -1,7 +1,7 @@
 package pluginapi_test
 
 import (
-	"io/ioutil"
+	"os"
 	"testing"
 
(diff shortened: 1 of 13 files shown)


cpoile (approved):
👍 Thanks Ben!

larkox (approved):
LGTM. The only small detail I would comment, is that not long ago, someone was having problems compiling one of these plugins/packages because they didn't have the correct version of go. Any way to avoid that?

hanzei:
@larkox Good suggestion. go1.16 is now required via `go.mod`.

PR #112, opened 2022-02-07
Flow improvements 

hanzei:
#### Summary
Targets: https://github.com/mattermost/mattermost-plugin-api/pull/111

- Added `GetCurrentStep` to API
- Removed `Do` and `Done` from API
- Fixed NPE if`handleDialog` returned an error map but no error
- `State` is now an `Map[string]interface{}` with some getterns
- Added `OnRender` option to step
- Added `WithColor` option to step
- If `forwardTo` is set, the next step will be automatically processed.  Use `""` to go to the next step.

#### Ticket Link
None

### DIFF
--- a/experimental/flow/flow.go
+++ b/experimental/flow/flow.go
@@ -86,6 +86,15 @@ func (f *Flow) ForUser(userID string) *Flow {
 	return &clone
 }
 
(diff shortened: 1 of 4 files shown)


levb (commented):
👍

PR #115, opened 2022-02-23
Flow: Step.WithImage improvement

levb:
#### Summary
`Step.WithImage`:
- No longer require pluginURL as parameter
- Accept full HTTP(s) URLs

#### Ticket Link
none

### DIFF
--- a/experimental/flow/step.go
+++ b/experimental/flow/step.go
@@ -83,7 +83,7 @@ func (s Step) Next(name Name) Step {
 
 func (s Step) WithImage(imageURL string) Step {
 	if u, err := url.Parse(imageURL); err == nil {
-		if u.Host != "" && (u.Scheme == "http" || u.Scheme == "https"){
+		if u.Host != "" && (u.Scheme == "http" || u.Scheme == "https") {
 			s.template.ImageURL = imageURL
 		} else {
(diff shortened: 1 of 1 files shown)


experimental/flow/step.go
```diff
@@ -80,9 +81,13 @@ func (s Step) Next(name Name) Step {
 	return s
 }
 
-func (s Step) WithImage(pluginURL, path string) Step {
-	if path != "" {
-		s.template.ImageURL = pluginURL + "/" + strings.TrimPrefix(path, "/")
+func (s Step) WithImage(imageURL string) Step {
+	if u, err := url.Parse(imageURL); err == nil {
```
hanzei, line 85: Nit: Should an error be logged if `err != nil`? Return one would make the API hard to use.

hanzei (approved):

experimental/flow/step.go
levb: Logging here is a bit hard, `Step` is a simple struct, doesn't have a logger. 1/5 just ignore invalid URLs... they'd look broken in the UI anyway, easy to spot.

hanzei: :+1:

PR #118, opened 2022-06-22
Improved logrus logging

lieut-data:
#### Summary
This change accompanies a Playbooks PR to improve the logging story with `logrus`:
* Set `logrus.TraceLevel` by default so that the server gets all logs and can do its own filtering. Plugins that forget to set a level end up with `logrus.InfoLevel` and have no way to programmatically enable debug logs.
* Add a `plugin_caller` field to logging events, tracing the logger from the plugin's perspective instead of having the server clobber it with a `caller` that's just `app/plugin_api.go`.

As part of this, I also eliminated the `errors` package, avoiding the circular dependency with `cluster` by just simplifying the version of `normalizeAppError` there.

#### Ticket Link
None.

### DIFF
--- a/cluster/job_once.go
+++ b/cluster/job_once.go
@@ -3,14 +3,11 @@ package cluster
 import (
 	"encoding/json"
 	"math/rand"
-	"net/http"
 	"sync"
 	"time"
 
 	"github.com/mattermost/mattermost-server/v6/model"
 	"github.com/pkg/errors"
-
-	apierrors "github.com/mattermost/mattermost-plugin-api/errors"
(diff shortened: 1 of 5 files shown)


cpoile (approved):
Thank you, we (Calls) should be using this instead of plain plugin logging. :)

lieut-data:
@cpoile, see also https://github.com/mattermost/mattermost-plugin-playbooks/pull/1271 for some other changes planned -- would love your feedback if interested.

mickmister (approved):

PR #119, opened 2022-07-29
go get github.com/mattermost/mattermost-server/v6@...

lieut-data:
#### Summary
Pin the mattermost-server dependency to a hash that includes `EnsureBotUser` from https://github.com/mattermost/mattermost-plugin-api/commit/0afd85e854d433c211773ee3c66152cc8e27f5ab.

#### Ticket Link
N/A


agnivade (approved):
Somehow golangci has gone bonkers in all repos. It was suddenly failing in the load test repo as well :man_shrugging:

lieut-data:
Ok, I messed with the golangci-lint pipeline and the go version and I /think/ everything passes now. Sorry for the churn!

go.mod
```diff
@@ -17,3 +17,65 @@ require (
 	golang.org/x/oauth2 v0.0.0-20211104180415-d3ed0bb246c8
 	golang.org/x/text v0.3.7
 )
+
+require (
```
lieut-data, line 21: I don't understand `go.mod` enough to know why this is suddenly necessary.

agnivade (approved):

go.mod
isacikgoz: Do you mean additional require block? If so;

>  Because the number of explicit requirements may be substantially larger in an expanded Go 1.17 go.mod file, the newly-added requirements on indirect dependencies in a go 1.17 module are maintained in a separate require block from the block containing direct dependencies. 

sauce: https://go.dev/doc/go1.17#Tools

isacikgoz (approved):

PR #121, opened 2022-09-20
Add build info command helper

hanzei:
#### Summary

Since go1.18 [`debug.BuildSetting`](https://pkg.go.dev/runtime/debug@go1.18beta2#BuildSetting) is provided as part of [`debug.BuildInfo`](https://pkg.go.dev/runtime/debug@go1.18beta2#BuildInfo). It contains build information like the build commit or build time. 
This PR adds a helper helper which allows them to easily create a `about` command for easier debugging, without any build flags.

Screenshot from 2022-02-11 11-29-12

An example usage on the plugin side can be found [here](https://github.com/mattermost/mattermost-plugin-github/commit/0e7ae53899bfc84ae342288e87e59dff6b36cab9).

#### Ticket Link
None

### DIFF
--- a/experimental/command/info.go
+++ b/experimental/command/info.go
@@ -1,0 +1,63 @@
+package command
+
+import (
+	"fmt"
+	"regexp"
+	"runtime/debug"
+	"strings"
+	"time"
+
+	"github.com/mattermost/mattermost-server/v6/model"
+	"github.com/pkg/errors"
+)
+
+func BuildInfoAutocomplete(cmd string) *model.AutocompleteData {
+	return model.NewAutocompleteData(cmd, "", "Display build info")
+}
+
+func BuildInfo(manifest model.Manifest, props map[string]any) (string, error) {
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		return "", errors.New("failed to read build info")
+	}
+
+	var (
+		revision      string
+		revisionShort string
+		buildTime     time.Time
+	)
+	for _, s := range info.Settings {
+		switch s.Key {
+		case "vcs.revision":
+			revision = s.Value
+			revisionShort = revision[0:7]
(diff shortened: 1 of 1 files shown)


agarciamontoro (commented):
This is super useful! I have a couple of questions:
1. What is the goal of the props parameter? If it's just printing key-value pairs, I'd say that the caller code can do that itself, and we can make the code here simpler.
2. What if instead of returning a string we pass a new `BuildData` struct with all the info collected, so that plugins can access/analyze/print that in the way they want? If we want, we can also add a separate "func FormatBuildData(data BuildData) string", which would come down to that last `fmt.Sprintf` in the code.

hanzei:
@agarciamontoro 
1. Good point, I've removed `props`. It doesn't provide values besides as the caller can append the build info themself.
2. While I like the idea of allowing plugin to use the build info as they want, it kind of bloats the API.  It feels like a re-implementation of `debug.ReadBuildInfo`, which is not what I have in mind for this PR. Also, `BuildInfo` doesn't really fit in a `command` package any longer, but I don't want to split the functionally over multiple packages.
You can see how the changes would look like in [`5bc49d0` (#121)](https://github.com/mattermost/mattermost-plugin-api/pull/121/commits/5bc49d0612d62d73b4b2163c782323dddec64801). Please let me know what you think. I'm so sure I like them.

agarciamontoro (approved):
I like this approach better, but I'm only 3/5. If you feel strongly about having a simpler, more opinionated, API, then we can go with that first and revisit later if needed, I'm also ok with that! Your call :)

hanzei:
Actually, I much prefer the opinionated, simpler API.  Reverting the latest change.

javaguirre (approved):
Seems very useful, LGTM!

PR #124, opened 2022-11-08
Allow storing arbitrary data in a job

hanzei:
#### Summary
This PR allows plugins to store arbitrary data as part of their scheduled jobs. `props` is stored as part of the `JobOnce` struct in the KV store.

I tried implementing the changes using generics for type safety, but that didn't work out well with the current way the code is structured, as `JobOnceScheduler` is a global singleton. 

The code should look like this:
```golang
type JobOnceScheduler[T any] struct {
[...]
}

var s *JobOnceScheduler[any]

// GetJobOnceScheduler returns a scheduler which is ready to have its callback set. Repeated
// calls will return the same scheduler.
func GetJobOnceScheduler[T any](pluginAPI JobPluginAPI) *JobOnceScheduler[T] {
	schedulerOnce.Do(func() {
		s = &JobOnceScheduler[T]{
			pluginAPI: pluginAPI,
			activeJobs: &syncedJobs{
				jobs: make(map[string]*JobOnce),
			},
			storedCallback: &syncedCallback{},
		}
	})
	return s
}
``` 
which doesn't work as `JobOnceScheduler[T]` can't be assigned to a `JobOnceScheduler[any]`.

If this PR needs type safety, which would be nice, we would need to get rid of the singleton. That could be done by prefixing the key of each scheduler, a pattern that is used in other packages of `pluginapi`.

#### Ticket Link
Needed for https://github.com/mattermost/mattermost-plugin-apps/pull/375


cluster/job_once_scheduler.go
```diff
@@ -124,18 +124,18 @@ func (s *JobOnceScheduler) ListScheduledJobs() ([]JobOnceMetadata, error) {
 }
 
 // ScheduleOnce creates a scheduled job that will run once. When the clock reaches runAt, the
-// callback will be called with key as the argument.
+// callback will be called with key and props as the argument.
```
levb, line 127: is it possible to move `props` to the last parameter for better (inline) typography when invoking?

levb (approved):
LGTM other than 1 nit

cluster/job_once_scheduler.go
hanzei: Good suggestion. Added in [f8b73bc](https://github.com/mattermost/mattermost-plugin-api/pull/124/commits/f8b73bc6795e54f407ce5aab7cd58ef003bb1d62).

hanzei:
@levb Heads up that I added some thoughts about using generics.

cluster/job_once_example_test.go
```diff
@@ -1,13 +1,15 @@
...
+	"log"
 	"time"
 
 	"github.com/mattermost/mattermost-server/v6/plugin"
 )
 
-func HandleJobOnceCalls(key string) {
+func HandleJobOnceCalls(key string, props any) {
 	if key == "the key i'm watching for" {
+		log.Println(props)
```
agarciamontoro, line 12: Do we want this print?

agarciamontoro (approved):
Looks good! Just one nit below. I'd also want @cpoile to take a look, since he worked on this initially and he'll have the best context to review the PR.

cpoile (commented):
This is interesting! I think we might need to think a second though before merging, just in case --

For a bit of context -- the reason we didn't include a props in the first version of this was that the key the job returns can be the key in the kvstore -- so if you want to store a "props" you would store them in the kvstore with that key and pull it out once the jobonce fires. Which means less data trapped in goroutines...

This PR might let people think they can put anything into `Props`, which could cause some problems if they _do_ put anything in there...

Also, I haven't looked at Go generics yet, but do we have a way (e.g., type bounds?) to ensure that `Props` is `json.Marshal` and `json.Unmarshal`-able?

cluster/job_once_example_test.go
hanzei: Yes, I found that useful as part of the example.

hanzei:
> This PR might let people think they can put anything into `Props`, which could cause some problems if they _do_ put anything in there...

It's worth noting that the KV store methods like `Set` (https://github.com/mattermost/mattermost-plugin-api/blob/24f6408ed7a6a6830e8c3060ba9c865dbbe40335/kv.go#L58) also don't do a good job in documenting what types can be passed as values.

Given that `encode/json.Marshal` is used in the KV store, the [documentation](https://pkg.go.dev/encoding/json#Marshal) does also apply here. `Marshal` returns a `UnsupportedTypeError` if a value type is unsupported. The code propagates that error all the way to the top. Hence, that's an error that should be spotted during development.

> Also, I haven't looked at Go generics yet, but do we have a way (e.g., type bounds?) to ensure that `Props` is `json.Marshal` and `json.Unmarshal`-able?

I can't think of a way. Generics in go don't allow recursive type parameters. Hence, `struct` as a type argument has to contain a finite list of fields.

cpoile:
> It's worth noting that the KV store methods like Set (...) also don't do a good job in documenting what types can be passed as values.

True, but I think the real issue is whether we want to put big chunks of data in the job once goroutines. This was meant to be lightweight. The way it's designed now, there can be an unlimited number of long running goroutines parked waiting for their timer to expire.

If we really want to attach data to the job once items, I think we need to redesign how the system schedules and parks its future events. I think @lieut-data had a good idea: we could use a priority queue and periodically poll the db to add soon-to-fire items to that queue. That way we would have only the next x minutes of items parked at any one time. 

If we moved to something like that, I would feel a lot better letting people attach arbitrary amounts of data. 

Or, alternatively, set a max number of bytes and enforce that at runtime?

hanzei:
@cpoile A priority queue is an awesome idea to reduce the memory consumption of the job scheduler. But it's a bit out of scope of this PR. I think it would make for a great HW ticket!

As part of this PR I've added a limit of 10k bytes for `props` in [e50c0fb](https://github.com/mattermost/mattermost-plugin-api/pull/124/commits/e50c0fb7f7d37b6ffca4712675ca32caefcbca15). That should prevent enormous, long-lasting go routines.

cpoile (approved):
Cool, I think that's a reasonable compromise since parked routines are about 4k. 

I really hope people don't fill these props up though, that would have unexpected results if they don't understand what's really happening (every one is parked until it runs).

PR #126, opened 2023-02-23
Don't panic if no driver is provided

hanzei:
#### Summary
If ` pluginapi.NewClient` is called with no driver i.e. `pluginapi.NewClient(p.API, nil)` any call to the DB store will panic. This PR changes the code to be more defensive and return an error for that case.

#### Ticket Link
None

### DIFF
--- a/store.go
+++ b/store.go
@@ -89,6 +89,10 @@ func (s *StoreService) initialize() error {
 		return nil
 	}
 
+	if s.driver == nil {
+		return errors.New("no db driver was provided")
+	}
+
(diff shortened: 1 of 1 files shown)


srkgupta (approved):
Thanks @hanzei. LGTM

trilopin (approved):
Nice defensive point.

CircleCI is broken, I've checked locally that lint and tests run ok :)

hanzei:
Yeah, I've reached out via https://community.mattermost.com/core/pl/todgj1ngab8kdypfwcwhag4z3a

PR #129, opened 2023-03-20
Mm 51400 enable disable tracking

trilopin:
### Enable/Disable
Add the ability to enable/disable telemetry trackers at pluginAPI.

The main goal is to get tracker/client creation together and avoid hook order race conditions that end up with trackers with null clients or flowmanagers with outdated trackers.

Related gihub plugin pr https://github.com/mattermost/mattermost-plugin-github/pull/659

Add `Enable` and `Disable` to the Tracker interface and implementations to the tracker struct.

All telemetry client & tracker initialization is done at onActivate, and OnConfigurationChange will only take care of config reload.

OnActivate
```go
	// Telemetry client
	p.telemetryClient, err = telemetry.NewRudderClient()
	if err != nil {
		p.API.LogWarn("Telemetry client not started", "error", err.Error())
		return
	}

	// Get config values
	p.tracker = telemetry.NewTracker(
		p.telemetryClient,
		p.API.GetDiagnosticId(),
		p.API.GetServerVersion(),
		Manifest.Id,
		Manifest.Version,
		"shortname",
		telemetry.NewTrackerConfig(p.API),
		logger.New(p.API),
	)
```

OnConfigurationChange
```go
	if p.tracker != nil {
		p.tracker.ReloadConfig(telemetry.NewTrackerConfig(p.API))
	}
```

OnDeactivate
```go
	if err := p.telemetryClient.Close(); err != nil {
		p.API.LogWarn("Telemetry client failed to close", "error", err.Error())
	}
```

### Debug at development stage
Opportunistic change to debug telemetry (without rudder access).

The current documentation has a logger as a parameter of NewTracker (at doc.go and func docstring) but was removed at some point. However, I see much value in passing a logger to debug telemetry events. 

It's a bit defensive, and won't debug anything if nil is passed, and actually, from the NewTrackerConfig I choose to pass a logger just when ServiceSettings.EnableDeveloper is true. 

Screenshot 2023-03-20 at 17 49 49

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-51400

### DIFF
--- a/experimental/telemetry/doc.go
+++ b/experimental/telemetry/doc.go
@@ -36,26 +36,36 @@
 //     p.API.LogWarn("telemetry client not started", "error", err.Error())
 //     }
 //     ...
-//     p.tracker = telemetry.NewTracker(p.telemetryClient, p.API.GetDiagnosticId(), p.API.GetServerVersion(), manifest.Id,
-//     manifest.Version, "pluginName", enableDiagnostics, logger)
-//     }
-//  3. Init and update the tracker on configuration change
-//     func (p *Plugin) OnConfigurationChange() error {
-//     ...
-//     if enableDiagnostics && p.tracker != nil {
-//     p.tracker.Enable()
-//     } else if !enableDiagnostics && p.tracker != nil {
-//     p.tracker.Disable()
-//     }
-//     }
-//  4. Close the client on plugin deactivate
-//     func (p *Plugin) OnDeactivate() error {
-//     if p.telemetryClient != nil {
-//     err := p.telemetryClient.Close()
-//     if err != nil {
-//     p.API.LogWarn("OnDeactivate: failed to close telemetryClient", "error", err.Error())
-//     }
-//     }
-//     return nil
-//     }
+//     p.tracker = telemetry.NewTracker(
+//     p.telemetryClient,
+//     p.API.GetDiagnosticId(),
+//     p.API.GetServerVersion(),
+//     Manifest.Id,
+//     Manifest.Version,
+//     "github",
+//     telemetry.NewTrackerConfig(p.API),
+//     logger.New(p.API)
+//
+// )
+//
+//	   p.tracker = telemetry.NewTracker(p.telemetryClient, p.API.GetDiagnosticId(), p.API.GetServerVersion(), manifest.Id,
+//	   manifest.Version, "pluginName", telemetry.TrackerConfig{EnabledTracking: enabledTracking, EnabledLogging: enabledDeveloper}, logger)
+//	   }
+//	3. Init and update the tracker on configuration change
+//	   func (p *Plugin) OnConfigurationChange() error {
+//	   ...
+//	   if p.tracker != nil {
+//		p.tracker.ReloadConfig(telemetry.NewTrackerConfig(p.API))
+//	   }
+//	   }
+//	4. Close the client on plugin deactivate
+//	   func (p *Plugin) OnDeactivate() error {
+//	   if p.telemetryClient != nil {
+//	   err := p.telemetryClient.Close()
+//	   if err != nil {
+//	   p.API.LogWarn("OnDeactivate: failed to close telemetryClient", "error", err.Error())
+//	   }
+//	   }
+//	   return nil
+//	   }
 package telemetry
--- a/experimental/telemetry/doc.go
+++ b/experimental/telemetry/doc.go
@@ -34,23 +34,23 @@
 //
 // 2. Start the telemetry client and tracker on plugin activate
 //
-//		func (p *Plugin) OnActivate() error {
-//		    p.telemetryClient, err = telemetry.NewRudderClient()
-//		    if err != nil {
-//		        p.API.LogWarn("telemetry client not started", "error", err.Error())
-//		    }
-//	        ...
-//	        p.tracker = telemetry.NewTracker(
-//	           p.telemetryClient,
-//	           p.API.GetDiagnosticId(),
-//	           p.API.GetServerVersion(),
-//	           Manifest.Id,
-//	           Manifest.Version,
-//	           "plugin_short_namame",
-//	           telemetry.NewTrackerConfig(p.API.GetConfig()),
-//	           logger.New(p.API)
-//	        )
+//	func (p *Plugin) OnActivate() error {
+//		p.telemetryClient, err = telemetry.NewRudderClient()
+//		if err != nil {
+//			p.API.LogWarn("telemetry client not started", "error", err.Error())
 //		}
+//		...
+//		p.tracker = telemetry.NewTracker(
+//			p.telemetryClient,
+//			p.API.GetDiagnosticId(),
+//			p.API.GetServerVersion(),
+//			Manifest.Id,
+//			Manifest.Version,
+//			"plugin_short_namame",
+//			telemetry.NewTrackerConfig(p.API.GetConfig()),
+//			logger.New(p.API)
+//		)
+//	}
 //
 // 3. Trigger tracker changes when configuration changes
 //
--- a/experimental/telemetry/doc.go
+++ b/experimental/telemetry/doc.go
@@ -35,13 +35,18 @@
 //     if err != nil {
 //     p.API.LogWarn("telemetry client not started", "error", err.Error())
 //     }
+//     ...
+//     p.tracker = telemetry.NewTracker(p.telemetryClient, p.API.GetDiagnosticId(), p.API.GetServerVersion(), manifest.Id,
+//     manifest.Version, "pluginName", enableDiagnostics, logger)
 //     }
 //  3. Init and update the tracker on configuration change
 //     func (p *Plugin) OnConfigurationChange() error {
 //     ...
-//     p.tracker = telemetry.NewTracker(p.telemetryClient, p.API.GetDiagnosticId(), p.API.GetServerVersion(), manifest.Id,
-//     manifest.Version, "pluginName", enableDiagnostics, logger)
-//     return nil
+//     if enableDiagnostics && p.tracker != nil {
+//     p.tracker.Enable()
+//     } else if !enableDiagnostics && p.tracker != nil {
+//     p.tracker.Disable()
+//     }
 //     }
 //  4. Close the client on plugin deactivate
 //     func (p *Plugin) OnDeactivate() error {
--- a/experimental/telemetry/doc.go
+++ b/experimental/telemetry/doc.go
@@ -26,10 +26,10 @@
 // 1. Add the new fields to the plugin
 //
 //	type Plugin struct {
-//	    plugin.MattermostPlugin
-//	    ...
-//	    telemetryClient telemetry.Client
-//	    tracker         telemetry.Tracker
+//		plugin.MattermostPlugin
+//		...
+//		telemetryClient telemetry.Client
+//		tracker         telemetry.Tracker
 //	}
 //
 // 2. Start the telemetry client and tracker on plugin activate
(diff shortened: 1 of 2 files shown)


experimental/telemetry/tracker.go
```diff
@@ -55,6 +62,7 @@ func NewTracker(
 	pluginVersion,
 	telemetryShortName string,
 	enableDiagnostics bool,
+	l logger.Logger,
```
trilopin, line 65: 2/5 on backwards compatibility.

What's the current strategy for that? Should I be more conservative?

m1lt0n: My take (0/5) is that breaking backward compatibility is ok, as it will make (if someone bumps their version of the plugin API) the plugin not compile, so it's immediately identifiable. As long as we have it in the release notes as a breaking change, it should be fine to have the logger available in the tracker.

m1lt0n (approved):
Great one!

experimental/telemetry/tracker.go
trilopin: great! I'm glad to know those boundaries

experimental/telemetry/tracker.go
```diff
@@ -67,19 +75,37 @@ func NewTracker(
...
+	if t.logger == nil {
+		return
+	}
+	t.logger.Debugf(message, args...)
+}
+
 func (t *tracker) TrackEvent(event string, properties map[string]interface{}) error {
+	event = t.telemetryShortName + "_" + event
 	if !t.enabled || t.client == nil {
+		t.debugf("Plugin telemetry event `%s` triggered, but not sent due to configuration", event)
```
hanzei, line 102:
```suggestion
		t.debugf("Plugin telemetry event `%s` tracked, but not sent due to configuration", event)
```

experimental/telemetry/doc.go
```diff
@@ -35,13 +35,18 @@
...
 //     func (p *Plugin) OnConfigurationChange() error {
 //     ...
-//     p.tracker = telemetry.NewTracker(p.telemetryClient, p.API.GetDiagnosticId(), p.API.GetServerVersion(), manifest.Id,
-//     manifest.Version, "pluginName", enableDiagnostics, logger)
-//     return nil
+//     if enableDiagnostics && p.tracker != nil {
+//     p.tracker.Enable()
+//     } else if !enableDiagnostics && p.tracker != nil {
+//     p.tracker.Disable()
+//     }
```
hanzei, line 49: 1/5 it's easier to read if 
```suggestion
//  if p.tracker != nil {
//      if enableDiagnostics {
//          p.tracker.Enable()
//      } else {
//          p.tracker.Disable()
//      }
//  }
```

hanzei (commented):
See also https://github.com/mattermost/mattermost-plugin-github/pull/659#pullrequestreview-1351273291

m1lt0n:
👍 on the reload config and newtrackerconfig changes. Approval still stands

experimental/telemetry/tracker.go
```diff
@@ -45,16 +75,17 @@ type tracker struct {
 // - pluginVersion: The plugin version.
 // - telemetryShortName: Short name for the plugin to use in telemetry. Used to avoid dot separated names like `com.company.pluginName`.
 // If a empty string is provided, it will use the pluginID.
-// - enableDiagnostics: Whether the system has enabled sending telemetry data. If false, the tracker will not track any event.
-// - l Logger: A logger to log any error related with the telemetry tracking.
+// - config: Whether the system has enabled sending telemetry data. If false, the tracker will not track any event.
+// - l Logger: A logger to debug event tracking and some important changes (it wont log if nil is passed as logger).
```
mickmister, line 79:
```suggestion
// - l Logger: A logger to debug event tracking and some important changes (it won't log if nil is passed as logger).
```

experimental/telemetry/tracker.go
```diff
@@ -66,20 +97,41 @@ func NewTracker(
...
+func (t *tracker) ReloadConfig(config TrackerConfig) {
+	if config.EnabledTracking != t.enabledTracking {
+		if config.EnabledTracking {
+			t.debugf("Enabling plugin telemetry")
+		} else {
+			t.debugf("Disabling plugin telemetry")
+		}
+		t.enabledTracking = config.EnabledTracking
+	}
+	t.enabledLogging = config.EnabledLogging
```
mickmister, line 115: Same 0/5 preference of newlines after blocks that have code in the same block scope

```suggestion
		}

		t.enabledTracking = config.EnabledTracking
	}

	t.enabledLogging = config.EnabledLogging
```

mickmister (approved):
LGTM, just some non-blocking suggestions

hanzei:
@trilopin Heads up that there are merge conflicts due to https://github.com/mattermost/mattermost-plugin-api/pull/130

experimental/telemetry/tracker.go
```diff
@@ -34,7 +64,9 @@ type tracker struct {
 	pluginID           string
 	pluginVersion      string
 	telemetryShortName string
-	enabled            bool
+	enabledTracking    bool
+	enabledLogging     bool
```
hanzei, line 68: 1/5 should a `tracker` include a `TrackerConfig` instead?

experimental/telemetry/tracker.go
```diff
@@ -2,14 +2,44 @@ package telemetry
...
+	"github.com/mattermost/mattermost-plugin-api/experimental/bot/logger"
 )
 
+type TrackerConfig struct {
+	EnabledTracking bool
+	EnabledLogging  bool
+}
+
+// NewTrackerConfig returns a new trackerConfig from the current values from the plugin API.
+func NewTrackerConfig(api plugin.API) TrackerConfig {
```
hanzei, line 17: Passing a `plugin.API` makes the method hard to use with a https://pkg.go.dev/github.com/mattermost/mattermost-plugin-api#Client. Should `NewTrackerConfig` accept a `model.Config` instead?

experimental/telemetry/tracker.go
```diff
@@ -66,20 +99,42 @@ func NewTracker(
...
+	if config.EnabledTracking != t.enabledTracking {
+		if config.EnabledTracking {
+			t.debugf("Enabling plugin telemetry")
+		} else {
+			t.debugf("Disabling plugin telemetry")
+		}
+	}
+
+	t.enabledTracking = config.EnabledTracking
+	t.enabledLogging = config.EnabledLogging
```
hanzei, line 118: These `ReloadConfig` calls might happen in quick succession. Do any precautionary measures need to be taken here?

hanzei (changes requested):

experimental/telemetry/tracker.go
```diff
@@ -66,20 +99,42 @@ func NewTracker(
...
 		pluginID:           pluginID,
 		pluginVersion:      pluginVersion,
-		enabled:            enableDiagnostics,
+		enabledLogging:     config.EnabledLogging,
+		enabledTracking:    config.EnabledTracking,
+		logger:             l,
 	}
 }
 
+func (t *tracker) ReloadConfig(config TrackerConfig) {
```
hanzei, line 108: Nit pick:

The plugin code currently calls this method like this:
```golang
	if p.tracker != nil {
		p.tracker.ReloadConfig(telemetry.NewTrackerConfig(p.API))
	}
```

If the code should support the case that `tracker` is `nil`, it could do so by returning early:
```suggestion
func (t *tracker) ReloadConfig(config TrackerConfig) {
	if t == nil {
		return
	}
```

experimental/telemetry/tracker.go
trilopin: do you have previous experiences of onConfigurationChange called in a burst 
?
are you thinking of adding a lock? 

are you concerned about quick changes or about concurrent calls?

experimental/telemetry/tracker.go
trilopin: I don't feel the pressure yet, but I'm not against that if you perceive it as more readable/solid

experimental/telemetry/tracker.go
trilopin: it's ok to me as long as we hide the complexity of the nil-zero-values logic inside this func. 👍🏽

experimental/telemetry/tracker.go
trilopin: I've tried it but it ends up in a panic.

I think that since we work with the Tracker interface, when the realtracker still doesn't exist there's no type to cast to, and therefore it's pure nil instead of typed-nil.


https://play.golang.com/p/FrES0m9IV6X

hanzei: Good catch. I missed that `NewTracker` returns an interface instead of a concrete type. :+1: 

A bit off-topic: Per the proverb ("Accept interfaces, return structs")[https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b] I'm wondering if we should kill the `Tracker` interface entirely. That would also solve the `nil` issue.

experimental/telemetry/doc.go
```diff
@@ -32,22 +32,33 @@
...
+//	           p.telemetryClient,
+//	           p.API.GetDiagnosticId(),
+//	           p.API.GetServerVersion(),
+//	           Manifest.Id,
+//	           Manifest.Version,
+//	           "plugin_short_namame",
+//	           telemetry.NewTrackerConfig(p.API.GetConfig()),
+//	           logger.New(p.API)
+//	        )
+//		}
```
hanzei, line 53: The formating seems off. Can you double check?

experimental/telemetry/tracker.go
hanzei: There is nothing we can do about the quick changes. If the config changes, so does the tracker.

My question is about the concurrent calls. When race conditions happen, can they cause issues?

experimental/telemetry/tracker.go
trilopin: good point and agree with the overall approach about structs/intertfaces you mentioned.

however, to continue changing this means to make public tracker struct + choose a new name to avoid conflict with interface (or skipping revive linting rule). I'd keep that change out of this PR unless you think it's crucial.

experimental/telemetry/tracker.go
trilopin: I feel like overthinking a bit this but here we go:
- if concurrent calls are made with the same config, I don't see any potential issue (aside from some double login)
- if concurrent calls are made with different configs, we could end up with part of the config updated from one call and part from the other. To be honest, I don't know if that scenario can happen, but if that's the case I'd add a lock to make the changes atomically.


Do you have something in mind?

experimental/telemetry/doc.go
```diff
@@ -32,22 +32,33 @@
...
+//			p.telemetryClient,
+//			p.API.GetDiagnosticId(),
+//			p.API.GetServerVersion(),
+//			Manifest.Id,
+//			Manifest.Version,
+//			"plugin_short_namame",
+//			telemetry.NewTrackerConfig(p.API.GetConfig()),
+//			logger.New(p.API)
+//		)
 //	}
```
hanzei, line 53: Actually, the indenting you are using there is correct, and the existing one is wrong. Here tabs are used correctly, instead of spaces. Would you be interested in updating also the other code example here?

experimental/telemetry/tracker.go
hanzei: I'm fine with keeping that conversation out of this PR. :ok_hand:

experimental/telemetry/tracker.go
hanzei: I agree that a look should protect the writes to ensure no partial writes happen.

hanzei (approved):

PR #134, opened 2023-06-19
Use github.com/mattermost/mattermost/server/public

hanzei:
#### Summary
https://github.com/mattermost/mattermost/pull/23345 moved most of the plugin code into the `public/` submodule. This PR makes use of it and updates the dependencies to make use of `github.com/mattermost/mattermost/server/public`. Unfortunately, it's still necessary to import `github.com/mattermost/mattermost/server/v8`.

This PR is more of a POW to spot any issues with the new `public` module. It should not be merged until v8.0 is imminent.

#### Ticket Link
None


configuration.go
```diff
@@ -70,12 +70,11 @@ func (c *ConfigurationService) CheckRequiredServerConfiguration(req *model.Confi
 
 	cfg := c.api.GetConfig()
 
-	mc, err := utils.Merge(cfg, req, nil)
+	mergedCfg, err := config.Merge(cfg, req, nil)
```
hanzei, line 73: Maybe it's worth moving https://github.com/mattermost/mattermost/blob/58a7679daa4f463b060e18728216e6b6889caeea/server/config/utils.go#L166 into `public/model` and making it a method of `model.Config`. That way the import of `github.com/mattermost/mattermost/server/v8/config"` becomes unnecessary.

store.go
```diff
@@ -8,8 +8,8 @@ import (
 	_ "github.com/go-sql-driver/mysql"
 	_ "github.com/lib/pq"
 
-	"github.com/mattermost/mattermost-server/v6/plugin"
-	"github.com/mattermost/mattermost-server/v6/shared/driver"
+	"github.com/mattermost/mattermost/server/public/plugin"
+	"github.com/mattermost/mattermost/server/v8/platform/shared/driver"
```
hanzei, line 12: With the removal of Playbooks and Board from the mono repo, `github.com/mattermost/mattermost/server/v8/platform/shared/drive` is not longer imported by any packages in the mono repo. Any objections to moving it into `public`?

lieut-data (approved):

hanzei:
Not longer needed, as the code now lives in the mono repo: https://github.com/mattermost/mattermost/pull/24235

